PR #3, opened 2022-09-01
Steps/PrGate.yml: Add pkg_count condition to build steps

makubacki:
Adds a condition to build steps to only execute if pkg_count is
greater than zero. The pkg_count is set to a default value of 1.
Stuart PR Eval might set it to zero if that is used and no packages
need to be built. When it does so, it also clears the package name
variable. So not only is the a performance optimization, it is
required if Stuart PR Eval runs and determines no packages need
to be built.

### DIFF
--- a/Steps/PrGate.yml
+++ b/Steps/PrGate.yml
@@ -84,7 +84,7 @@ steps:
     inputs:
       filename: stuart_ci_setup
       arguments: -c ${{ parameters.build_file }} -p $(pkgs_to_build) --force-git -t ${{ parameters.build_targets}} -a ${{ parameters.build_archs}} TOOL_CHAIN_TAG=${{ parameters.tool_chain_tag}}
-    condition: succeeded()
+    condition: and(gt(variables.pkg_count, 0), succeeded())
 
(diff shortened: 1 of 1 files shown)


Steps/PrGate.yml
```diff
@@ -84,44 +84,44 @@ steps:
...
       arguments: -c ${{ parameters.build_file }} -p $(pkgs_to_build) -t ${{ parameters.build_targets}} -a ${{ parameters.build_archs}} TOOL_CHAIN_TAG=${{ parameters.tool_chain_tag}}
-    condition: succeeded()
+    condition: and(gt(variables.pkg_count, 0), succeeded())
 
 # Publish Test Results to Azure Pipelines/TFS
 - task: PublishTestResults@2
   displayName: Publish junit Test Results
   continueOnError: true
-  condition: succeededOrFailed()
+  condition: and(gt(variables.pkg_count, 0), succeeded())
```
makubacki, line 124: I'm going to move these log conditions back to `succeededOrFailed()`

makubacki: Done

cfernald (approved):

PR #5, opened 2022-09-02
Add CodeQL reusable GitHub workflow

makubacki:
Adds a new workflow YAML file that can be reused by repos that need
to run CodeQL.

The workflow accepts the following input strings:
  - setup_command - The (Stuart) setup command to run
  - update_command - The (Stuart) update command to run
  - build_command - The (Stuart) build command to run
  - python_version - The Python version to use

The command strings can be an empty string to skip that command.

The default Python version is currently '3.x'.

This reusable workflow encapsulates common CodeQL pre and post
build steps and centralizes default values for common configuration
points like the Python version to use for CodeQL workflows.

### DIFF
--- a/.github/workflows/CodeQl.yml
+++ b/.github/workflows/CodeQl.yml
@@ -1,0 +1,88 @@
+## @file
+# GitHub action CodeQL reusable workflow file.
+#
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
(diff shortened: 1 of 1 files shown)


makubacki:
Here's an example of a successful CodeQL action in mu_feature_mm_supv using this workflow: https://github.com/microsoft/mu_feature_mm_supv/runs/8163325222?check_suite_focus=true

kuqin12 (approved):

apop5 (approved):

PR #15, opened 2022-10-21
InstallMarkdownLint.yml: Update to markdownlint-cli 0.31.1

makubacki:
Updates to the latest markdownlint-cli to stay current. 0.31.1 was released
on 2/9/2022 per the version history:

https://www.npmjs.com/package/markdownlint-cli

Note: 0.32 introduces new failures in some repos that would need to be
addressed before moving to 0.32.x. 0.32.2 was specifically tested.

### DIFF
--- a/Steps/InstallMarkdownLint.yml
+++ b/Steps/InstallMarkdownLint.yml
@@ -7,6 +7,6 @@
 
 steps:
 
-- script: npm install -g markdownlint-cli@0.31.0
+- script: npm install -g markdownlint-cli@0.31.1
   displayName: Install Markdown Linter
   condition: succeeded()


apop5:
Has this script been triggered on any of the other repos? Or, by updating this, we are going to cause failures for the next PRs on the other repos?

makubacki:
> Has this script been triggered on any of the other repos? Or, by updating this, we are going to cause failures for the next PRs on the other repos?

It should not. We were already using 0.31.1 in mu_basecore. Changes across the same minor version are generally compatible (0.31 in this case). I will check on some others though because this and spell check are especially sensitive.

I tried 0.32 and I know that introduces new failures. That's why I'm keeping us 0.31.1 right now.

apop5 (approved):

cfernald (approved):

PR #18, opened 2022-10-21
Simplify job names

makubacki:
Removes parameterized input from job names so the name is not
dependent on variable resolution.


makubacki:
This is basically needed so variables can be passed as tool chain tags.

The job name does not really need the tool chain tag in it.

apop5 (approved):

apop5 (approved):

kuqin12 (approved):

PR #20, opened 2022-10-24
Update code to align with previous isOutput bug behavior

makubacki:
Since "isOutput" was previously spelled incorrectly (see d326659), the resulting
behavior was equivalent to "isOutput" not being specified at all.

Subsequent code has been written such that it refers to the variable
as a variable set in the same job rather than a multi-job variable as
"isOutput" is used for - the code simply refers to the variable name
rather than \<step\>.\<var name\>.

This change simply removes "isOutput" so the original set of code
assumptions continue to work as before.

### DIFF
--- a/Steps/PrGate.yml
+++ b/Steps/PrGate.yml
@@ -80,7 +80,7 @@ steps:
     inputs:
       filename: stuart_pr_eval
       # Workaround an azure pipelines bug.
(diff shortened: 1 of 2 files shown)


Steps/PrGate.yml
```diff
@@ -80,7 +80,7 @@ steps:
     inputs:
       filename: stuart_pr_eval
       # Workaround an azure pipelines bug.
-      arguments: -c ${{ parameters.build_file }} -p ${{ parameters.build_pkgs }} --pr-target origin/$(System.PullRequest.targetBranch) --output-csv-format-string "##vso[task.setvariable variable=pkgs_to_build;isOutpout=true]{pkgcsv}" --output-count-format-string "##vso[task.setvariable variable=pkg_count;isOutput=true]{pkgcount}"
```
makubacki, line 83: I forgot to push the commit that fixed this instance of `isOutpout` in #19 .

apop5 (approved):

PR #21, opened 2022-10-25
Add support for platform build using containers

cfernald:
- Remove need for path workaround
- Add separate pip install for Linux
- Add install_tools parameter for BuildPlatform

### DIFF
--- a/Steps/SetupPythonPreReqs.yml
+++ b/Steps/SetupPythonPreReqs.yml
@@ -23,6 +23,11 @@ steps:
       versionSpec: ">=3.10.6"
       architecture: x64
 
-- script: pip install ${{ parameters.pip_requirement_files }} --upgrade
-  displayName: Install and Upgrade pip Modules
-  condition: succeeded()
+- ${{ if ne(variables['Agent.OS'], 'Linux') }}:
+  - script: sudo pip install ${{ parameters.pip_requirement_files }} --upgrade
+    displayName: Install and Upgrade pip Modules
+    condition: succeeded()
+- ${{ else }}:
+  - script: pip install ${{ parameters.pip_requirement_files }} --upgrade
+    displayName: Install and Upgrade pip Modules
+    condition: succeeded()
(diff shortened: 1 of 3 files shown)


kuqin12 (approved):

Steps/SetupPythonPreReqs.yml
```diff
@@ -23,6 +23,10 @@ steps:
       versionSpec: ">=3.10.6"
       architecture: x64
```
cfernald, line 25: @makubacki & others. I'm not a huge fan of this solution so I'm open to other suggestions on this one. I'm trying to remove the hacky path work-around by running as root. The problem is I haven't found a clean way of using sudo based on the agent OS without a condition and two separate steps. Any ideas?

makubacki: I suppose making a virtual environment is cross-platform and would likely allow you to drop sudo. Conceptually, I think removing sudo is nice. However, the path to activate the virtual environment differs between Linux and Windows so I'm not sure it helps the OS check.

I don't think the OS check is that bad if just limited to the condition for these small tasks here. It's one of those things I think we really need to question (like you're doing) when we have to use it.

os-d (approved):

makubacki (approved):

Steps/SetupPythonPreReqs.yml
cfernald: Yeah, of all the way I explored this is the cleanest, it just adds that no-op step. The only viable alternative I could think of that's not too bad would be add a parameter for agent OS and use that to setup a prefix variable to make this a single operation, but that feels like just pushing the complexity to the consumer to remove that no-op step.

Ultimately, the issue is that pip doesn't provide a way that i've been able to find to install a package to the default system location when running as a non-root user in linux without explicit call outs to the OS environment like path or users.

cfernald: If there are no concerns i'll probably leave this as for now unless I can find a better solution to the user issue.

makubacki: I think it is fine to leave as-is for now

PR #22, opened 2022-11-15
Revert container updates in 5f3c041

makubacki:
Moves pip installation behavior back to what was done before
5f3c041. After analyzing several options, including using virtual
environments, this is the path that most consistently and simply
works for container and non-container builds.

It could potentially be built upon in the future, but it gets all of
the builds functional for now.

Resolves the following warning currently issued by Linux build
agents since 5f3c041 was committed:

"WARNING: Running pip as the 'root' user can result in broken
permissions and conflicting behaviour with the system package
manager.

It is recommended to use a virtual environment instead:
https://pip.pypa.io/warnings/venv"

### DIFF
--- a/Steps/SetupPythonPreReqs.yml
+++ b/Steps/SetupPythonPreReqs.yml
@@ -23,7 +23,13 @@ steps:
       versionSpec: ">=3.10.6"
       architecture: x64
 
-- script: sudo pip install ${{ parameters.pip_requirement_files }} --upgrade
+- script: |
+    python3 -m venv .venv
+    source .venv/bin/activate
+  displayName: Setup Python Virtual Environment (Linux)
+  condition: and(succeeded(), eq(variables['Agent.OS'], 'Linux'))
+
+- script: pip install ${{ parameters.pip_requirement_files }} --upgrade
   displayName: Install and Upgrade pip Modules (Linux)
   condition: and(succeeded(), eq(variables['Agent.OS'], 'Linux'))
 
--- a/Steps/SetupPythonPreReqs.yml
+++ b/Steps/SetupPythonPreReqs.yml
@@ -23,12 +23,6 @@ steps:
       versionSpec: ">=3.10.6"
       architecture: x64
 
-- script: |
-    python3 -m venv .venv
-    source .venv/bin/activate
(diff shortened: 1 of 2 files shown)


makubacki:
Note: I chose not to use a virtual environment on Windows because there's some known caveats there, example: https://faun.pub/azure-pipelines-and-python-virtual-environments-fdc1089656db

And it's not an issue on Windows agents at the moment.

Steps/SetupPythonPreReqs.yml
```diff
@@ -23,7 +23,13 @@ steps:
...
-- script: sudo pip install ${{ parameters.pip_requirement_files }} --upgrade
+- script: |
+    python3 -m venv .venv
+    source .venv/bin/activate
+  displayName: Setup Python Virtual Environment (Linux)
+  condition: and(succeeded(), eq(variables['Agent.OS'], 'Linux'))
+
+- script: pip install ${{ parameters.pip_requirement_files }} --upgrade
   displayName: Install and Upgrade pip Modules (Linux)
   condition: and(succeeded(), eq(variables['Agent.OS'], 'Linux'))
```
cfernald, line 34: We might as well drop the separate Linux step then right?

makubacki: I prefer them separate so we can better isolate which operation failed (1) creating the virtual environment or (2) install the pip modules in the virtual environment.

cfernald: I meant the "Install and Upgrade pip Modules" since it is now identical to the windows step below

makubacki: Sorry, misunderstood. Yeah, I'll merge those.

cfernald (approved):

Steps/SetupPythonPreReqs.yml
makubacki: Done in this PR commit https://github.com/microsoft/mu_devops/pull/22/commits/42f505f868835945bda5e96379061187cdb9bedc

spbrogan (approved):

kuqin12 (approved):

cfernald (approved):

PR #40, opened 2022-11-30
Add NuGet publishing

makubacki:
Introduces the ability to publish NuGet package releases from
Project Mu repos.

Also includes changes to publish artifacts by type to more cleanly
control how they are produced & consumed - binaries, logs, and other.

**Integration Notes**

This commit updates the major version of mu_devops. This means it
might require integration work that will result in a repo build
breaking unless performed.

- Artifacts are now published under three separate categories of "binaries",
- "logs", and "other"
  - Previously all artifacts were published under `"Build Logs $(System.JobName)"`
  - Any flows dependent on artifact names will need to be updated
  - Three new templates are provided for reusable publishing of content in these categories:
    1. `Steps/BinaryCopyAndPublish.yml`
    2. `Steps/CommonLogCopyAndPublish.yml`
    3. `Steps/OtherCopyAndPublish.yml`
- `Steps/PrGate.yml` has a new template parameter - `artifacts_identifier`
  - This can be used to adjust the name assigned to artifacts so it makes the
    most sense for a given platform
    - The default value is an empty string
    - For most platforms, it is recommended to pass the package name and build target
- `Steps/PrGate.yml` has new template parameters to control the binary and other
  content published.
  - The default value for both is an empty string

**Non-Breaking Change Notes**

- There is a new step template to easily publish content from the three artifact
  categories of a given pipeline to NuGet (`Steps/NuGet.yml`).
  - It is a step template so it can easily access file content already on the job
    build agent.
  - It provides the ability to select which categories of artifacts are published.
- There is a new job template to generate a build matrix. This is tailored toward
  firmware build scenarios that involve groups of packages that are built
  together on a single agent and others that are individually built on a dedicated agent.

### DIFF
--- a/Steps/NuGet.yml
+++ b/Steps/NuGet.yml
@@ -1,0 +1,165 @@
+## @file
+# Mu DevOps template file to build and publish Nuget packages.
+#
+# Specify the NuGet feed API key as a secret pipeline variable named
+# "NUGET_KEY" if an API key is used.
+#
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+parameters:
+- name: artifacts_identifier
+  displayName: Artifacts Identifier
+  type: string
+  default: 'Artifacts'
+- name: custom_nuget_prep_step_list
+  displayName: Custom NuGet Package Step List
+  type: stepList
+  default:
+    - script: echo No custom NuGet prep steps provided
+- name: include_binaries
+  displayName: Include build binaries in the Nuget package
+  type: boolean
+  default: true
+- name: include_logs
+  displayName: Include build logs in the Nuget package
+  type: boolean
+  default: false
+- name: include_other
+  displayName: Include other build artifacts in the Nuget package
+  type: boolean
+  default: false
+- name: license_file_path
+  displayName: License File Path (Absolute Path)
+  type: string
+  default: ''
+
+#
+# Nuget Package Configuration Data
+#
+
+# The package config file path is relative to the source code workspace root directory.
+- name: nuget_package_config_file_path
+  displayName: Nuget Package Config File Path
+  type: string
+  default: ''
+
+# The package version should always be specified and accurate.
+- name: nuget_package_version
+  displayName: Nuget Package Version
+  type: string
+  default: ''
+
+steps:
+  - bash: |
+      if [ -z "${{ parameters.license_file_path }}" ]; then
+        echo "##vso[task.logissue type=error;]Missing template parameter \"license_file_path\""
+        echo "##vso[task.complete result=Failed;]"
+      fi
+      if ![ -f "${{ parameters.license_file_path }}" ]; then
+        echo "##vso[task.logissue type=error;]license_file_path does not exist: \"${{ parameters.license_file_path }}\""
+        echo "##vso[task.complete result=Failed;]"
+      fi
+    displayName: Validate License File Path
+
+  - bash: |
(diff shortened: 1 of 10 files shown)


makubacki:
In draft while still testing

makubacki:
Mostly done. Some minor changes will be made to the NuGet.yml template before the PR is published.

Needs https://github.com/tianocore/containers/issues/51 completed to bring Powershell into the Fedora container image being used in some builds.

makubacki:
Changes are done. Might refactor the NuGet file parameters in the future.

Steps/NuGet.yml
```diff
@@ -0,0 +1,159 @@
...
+    displayName: Validate License File Path
+
+  - bash: |
+     ARTIFACTS_IDENTIFIER_PATH=$(echo "${{ parameters.artifacts_identifier }}" | tr " " "_")
+     echo "##vso[task.setvariable variable=artifacts_identifier_path]$ARTIFACTS_IDENTIFIER_PATH"
+
+  #
+  # Binary Artifact Download
+  #
+  # The default steps download the binary build artifacts for the current build
```
cfernald, line 70: I'm still wrapping my head around the matrix system, but would this limit us to only being able to publish from a given OS build type per nuget package? E.g. a given nuget package which is from a given build could only contain the output of a VS or GCC build but not both?

makubacki: This file publishes one of three categories of files from the current build:

1. Binaries
2. Logs
3. Other

As far as this file is concerned, anything could be in the artifacts within these three categories. That's one of the reasons I chose to use "Artifact Identifier" instead of something more specific like "build type/target/OS/etc."

This change does break out these three categories within Mu DevOps (NuGet aside) to more easily track production and consumption of content in these categories, both for direct download/inspection and usage in post-steps like NuGet publishing.

A matrix could definitely drive calls to the NuGet.yml template, but it's not strictly required.

For example, in the [test change](https://github.com/makubacki/mu_tiano_platforms/commit/285b9d5a37753cd2f2d441812b9a1a88d8b15bf6#diff-3bcb307b8ffba6021ef8b268c86c9975d261e2738507169c1e0fab7763d65e67) in mu_tiano_platforms, the `artifacts_identifier` connects what artifacts are published in the `Steps/BuildPlatform.yml@mu_devops` template to what is consumed in the `Steps/NuGet.yml@mu_devops` template. That happens to be driven via a simple template, but it does not need to be. Also, any template or artifact push could be substituted instead of `Steps/BuildPlatform.yml@mu_devops` and as long as `artifacts_identifer` is set correctly for `Steps/NuGet.yml@mu_devops`, that content will be picked up and published to the NuGet package.

Since, in my test example, the same matrix output path is used for both the `BuildPlatform.yml` call and the `NuGet.yml` call, that means that each target/OS would be published separately.

To combine them, we would need to publish everything needed as one set of artifacts and publish that with an `artifacts_identifier` value passed to `NuGet.yml`.

---

The NuGet.yml interface is meant to provide a fairly generic way to pass content in. There's room to refactor the way content is published and/or extend what is there. For example, to do something like combine all OS builds into one NuGet package. In that case, a new pipeline could discover separate pre-existing OS builds of a commit (with something like a "release" label applied to them), publish them as a single set of build artifacts, and then call `NuGet.yml` with the `artifacts_identifier` value for what it collected/published.

Given build types supported is an individual repo detail, I was thinking this collecting of content across various builds would be done there and then `NuGet.yml` called with the `artifacts_identifier` value.

I was going to update my mu_tiano_platforms change to integrate this. I'm happy to do something like that if a single NuGet package is preferred.

apop5 (approved):

cfernald (approved):

PR #41, opened 2022-11-30
.sync/markdownlint.yaml: Update brace syntax

makubacki:
The text already comes with braces, so they are not needed explicitly
in the YAML config file.


makubacki:
Submitting as a fix to https://github.com/microsoft/mu_devops/pull/38

PR #45, opened 2022-12-01
Fix Mu DevOps Version Used on Release Version [Rebase & FF]

### LINKED ISSUE #44: Update dependencies on mu_devops to use release versions
Dependencies on the mu_devops repo currently use `mu_devops/main`.

To improve traceability and stability, they should use a release version of the repo.

makubacki:
Closes #44

This series makes two main changes:

1. Fixes all files dependent on the Mu DevOps repo to a fixed (release) version
   of the repo. The version is tracked in a single location in source control.
2. Adds Azure Pipeline file sync for all Project Mu repos originally left out of
   the original file sync change because they had custom build steps.

For reference, the v0.1.0 release of mu_devops is here:
https://github.com/microsoft/mu_devops/releases/tag/v0.1.0

Additional details for each commit are below...

---

## Files.yml: Add remaining Mu repo pipelines to file sync

Some repos were left out of the original file sync change because
they had custom build steps. Those pipeline files need to be in
Mu DevOps to consolidate various versioning and other details in
the file and it is also more straightforward to follow a process
where all pipeline files are simply maintained in the same
repo/place.

This change adds each repo's file as a unique file in the `.sync`
directory. While some level of commonality could be achieved with
templating, that would likely introduce unnecessary complexity.

These files can have the flexibility to customize most changes
directly in the file without having to worry about adjusting
template input that affects other custom step YAML files. Breaking
out the files like this also allows separate file history tracking
which is beneficial in 1:1 file to repo mapping.

Here the files map from their directory to their repo and the files
can pick up any common Mu DevOps customization needed before they
are synced.

---

## .sync: Fix Mu DevOps version in Mu pipelines

Project Mu pipelines use pipelines templates as a repository resource
from mu_devops. Currently, a pipeline build pulls the latest commit
from the `main` branch.

This means the version of mu_devops is "floating" in that pipelines
immediately pick up the latest change. To improve build traceability
and stability, this change moves the "leaf" pipeline YAML files to
use fixed/release versions of content in the mu_devops repo.

This change updates all pipeline files to use the version specified
in one place - `.sync/Version.njk` during the file sync process to
downstream repos.

Example flow:
  1. Make a new Mu DevOps release (e.g. "v1.0.0")
  2. Update `.sync/Version.njk` to set `mu_devops` to `"v1.0.0"`

Step (2) causes the following automated actions to take place after
the change is merged into `mu_devops/main`:
  1. All sync files (e.g. an Azure Pipeline file) that depend on
     this version get the new value substituted
  2. All sync files with the substituted value are synced to their
     respective repos (PRs created with the change)
  3. After the PRs in those repos are merged, they use the new
     version of Mu DevOps in their pipelines

---

## .sync: Fix Mu DevOps version in Mu workflows

Project Mu repos have "leaf" workflows. These are workflow files
that wrap around reusable workflows in mu_devops.

Currently, a leaf workflow uses the reusable workflow from the latest
commit in `mu_devops/main`.

This means the version of mu_devops is "floating" in that the leaf
workflows immediately pick up the latest change. To improve build
traceability and stability, this change moves the "leaf" workflow
files to use fixed/release versions of content in the mu_devops repo.

This change updates all leaf workflow files to use the version
specified in one place - `.sync/Version.njk`. That version will be
substituted in the leaf workflow file during the file sync process to
downstream repos.

Example flow:
  1. Make a new Mu DevOps release (e.g. "v1.0.0")
  2. Update `.sync/Version.njk` to set `mu_devops` to `"v1.0.0"`

Step (2) causes the following automated actions to take place after
the change is merged into `mu_devops/main`:
  1. All sync files (e.g. a leaf workflow file) that depend on
     this version get the new value substituted
  2. All sync files with the substituted value are synced to their
     respective repos (PRs created with the change)
  3. After the PRs in those repos are merged, they use the new
     version of Mu DevOps in their pipelines


apop5 (approved):

kenlautner (approved):

apop5:
I have some concerns about all of the custom azure pipelines.

Are we going to create more work by moving the customizations in this repo?  Could the same effect be accomplished without the need to have specific

makubacki:
> I have some concerns about all of the custom azure pipelines.
> 
> Are we going to create more work by moving the customizations in this repo? Could the same effect be accomplished without the need to have specific files for each repo in mu_devops?

We could likely consolidate the pipelines in the "custom" folder. Getting everything in order is a series of incremental improvements:

- Original: Random files all over, no shared infrastructure
- Next: Manually make infrastructure copied and pasted in repos as similar as possible
- Next: Store common config settings in single pipeline variable groups instead of spread over files across repos
- Next: Create Mu DevOps and consolidate most pipeline logic - wrap it with minimal YAML files in repos
- Next: Move minimal pipelines exactly the same (those in the "matrix_dependent" directory here) into a single set of files
- **Now**: Collect all "not exactly the same wrappers" in one place
- Next: Consolidate those as much as possible - templatize differences where it makes sense

We could make further improvements from there. But I think we're getting close to something reasonable.

PR #59, opened 2022-12-05
Add container logic for matric builds and enable containers for other.s

cfernald:
- Add logic to use container build for GCC matrix dependent builds
- Enable container build on all GCC CI pipelines
- Add devcontainer file to file sync

### DIFF
--- a/.sync/Files.yml
+++ b/.sync/Files.yml
@@ -366,3 +366,20 @@ group:
       microsoft/mu_silicon_intel_tiano
       microsoft/mu_tiano_platforms
       microsoft/mu_tiano_plus
+
+# Development Container - Common for CI based repos
+  - files:
+    - source: .sync/devcontainer/devcontainer.json
+      dest: .devcontainer/devcontainer.json
+    repos: |
+      microsoft/mu_basecore
+      microsoft/mu_common_intel_min_platform
+      microsoft/mu_crypto_release
+      microsoft/mu_feature_config
+      microsoft/mu_feature_ipmi
+      microsoft/mu_feature_uefi_variable
+      microsoft/mu_oem_sample
+      microsoft/mu_plus
+      microsoft/mu_silicon_arm_tiano
+      microsoft/mu_silicon_intel_tiano
+      microsoft/mu_tiano_plus
--- a/.sync/azure_pipelines/matrix_dependent/Ubuntu-GCC5.yml
+++ b/.sync/azure_pipelines/matrix_dependent/Ubuntu-GCC5.yml
@@ -23,6 +23,9 @@ resources:
       endpoint: microsoft
       name: microsoft/mu_devops
       ref: refs/tags/{{ sync_version.mu_devops }}
+  containers:
+    - container: linux-gcc
+      image: ghcr.io/tianocore/containers/fedora-35-build:2113a0e
 
 variables:
 - group: architectures-arm-64-x86-64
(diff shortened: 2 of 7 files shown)


cfernald:
Leaving this a draft until the dependent PRs are closed.

Doing basecore first and then will open the rest after feedback there: https://github.com/microsoft/mu_basecore/pull/199/files

os-d (approved):

.sync/azure_pipelines/matrix_dependent/Ubuntu-GCC5.yml
```diff
@@ -23,6 +23,9 @@ resources:
       endpoint: microsoft
       name: microsoft/mu_devops
       ref: refs/tags/{{ sync_version.mu_devops }}
+  containers:
+    - container: linux-gcc
+      image: ghcr.io/tianocore/containers/fedora-35-build:2113a0e
```
apop5, line 28: nit: This is calls for a fedora 35 but the name of the file is Ubuntu....

apop5 (approved):

.sync/azure_pipelines/matrix_dependent/Ubuntu-GCC5.yml
cfernald: Yeah, I think using the container we should probably distance ourselves some a specific SKU in the naming. I want to go through and rename all such files to Linux-GCC.yml. I will likely do this after some of the dust settles a bit as it will require editing all the pipeline configurations as well.

makubacki: I don't think it really matters since the VM images are Ubuntu and likely will be for the foreseeable future. As long as the VM image and the container image names line up, I don't see a need for extra repo thrash.

makubacki, line 28: I think it would be nice to define the container in one place in the repo.

I planned for [.sync/Version.njk](https://github.com/microsoft/mu_devops/blob/main/.sync/Version.njk) to contain versions of various resources used in files that are synced. What do you think about adding something like a `linux_gcc_container` entry there?

The file is already imported here so you would just need to use:

```yaml
  containers:
    - container: linux-gcc
      image: {{ sync_version.linux_gcc_container }}
```

.sync/devcontainer/devcontainer.json
```diff
@@ -0,0 +1,12 @@
+{
+  "image": "ghcr.io/tianocore/containers/fedora-35-dev:latest",
```
makubacki, line 2: If we want to keep the devcontainer on the latest image (which I think is fine), we would add another entry to `.sync/Version.njk` like `linux_gcc_container_dev` and set it to this then that is centralized as well.

.sync/azure_pipelines/matrix_dependent/Ubuntu-GCC5.yml
cfernald: in fact, it has to be ubuntu for the pipeline container support from the documentation. Either way, we can continue this discussion offline but i'll resolve this thread for now.

cfernald: I agree this is better. Unfortunately, the other referencing location is currently in a non file-synced file so I fear the returns may not be that great for the time being.

.sync/devcontainer/devcontainer.json
cfernald: I think this one if fine to leave locally defined since its intended for local development and is the only place where it will be used so far so I think I may leave it for a sake of readability. We wont be changing it much anyways since the its using latest, which I feel just avoids churn for a less critical versioning.

.sync/Files.yml
```diff
@@ -375,3 +375,20 @@ group:
...
+  - files:
+    - source: .sync/devcontainer/devcontainer.json
+      dest: .devcontainer/devcontainer.json
+    repos: |
+      microsoft/mu_basecore
+      microsoft/mu_common_intel_min_platform
+      microsoft/mu_crypto_release
+      microsoft/mu_feature_config
+      microsoft/mu_feature_ipmi
+      microsoft/mu_feature_uefi_variable
```
kuqin12, line 389: Add mm_supv?

kuqin12 (approved):

PR #76, opened 2023-01-07
Split approval and merge workflows

makubacki:
Split approval and merge workflows

Splitting the workflows provides several benefits:

1. Simplifies logic due to jobs being attached to separate triggers
   relevant for the specific job.
2. Reduces number of status checks shown in PRs due to certain steps
   being skipped in irrelevant circumstances.
3. Allows better workflow reuse.

The reviewers are also dropped from dependabot PRs since the pending
team reviewer is (1) not acted upon (2) can delay PR merging if the
reviewer team has not reviewed the PR.

The version is updated in anticipation of this change. It will be
modified if the expected version changes before this PR is merged.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -32,8 +32,6 @@ updates:
       prefix: "GitHub Action"
     labels:
       - "type:dependencies"
-    reviewers:
-      - "microsoft/project-mu-dependency-reviewers"
 
   - package-ecosystem: "pip"
     directory: "/"
@@ -44,5 +42,3 @@ updates:
     labels:
       - "language:python"
       - "type:dependencies"
-    reviewers:
-      - "microsoft/project-mu-dependency-reviewers"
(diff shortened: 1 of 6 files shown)


makubacki:
These are some adjustments I'd like to make after looking at some of the recent bot PRs. I'm planning to keep most of the bot PRs open until this is completed to run them against this set of changes.

.github/dependabot.yml
```diff
@@ -32,8 +32,6 @@ updates:
       prefix: "GitHub Action"
     labels:
       - "type:dependencies"
-    reviewers:
```
apop5, line 35: Taking the reviewers will reduce the number of notifications. Is there a way to notify reviewer if the CI fails so that manual action can be taken?

apop5 (approved):

.github/dependabot.yml
makubacki: There might be a better way, but we should be able to find them quick enough by checking for failed PRs from dependabot or uefibot.

Flickdm (approved):

TaylorBeebe (approved):

PR #87, opened 2023-01-17
Add dashboard for issue triage

apop5:
Added a issue triage dashboard that enables queries for no assignee and no labels

### DIFF
--- a/Notebooks/OpenIssues.github-issues
+++ b/Notebooks/OpenIssues.github-issues
@@ -38,5 +38,15 @@
     "kind": 2,
     "language": "github-issues",
     "value": "$repos is:open is:issue archived:false no:assignee"
+  },
+  {
+    "kind": 1,
+    "language": "markdown",
+    "value": "All Open Issues marked stale"
+  },
+  {
+    "kind": 2,
+    "language": "github-issues",
+    "value": "$repos is:open is:issue archived:false label:state:stale"
   }
 ]
--- a/Notebooks/OpenIssues.github-issues
+++ b/Notebooks/OpenIssues.github-issues
@@ -1,0 +1,42 @@
+[
+  {
+    "kind": 1,
+    "language": "markdown",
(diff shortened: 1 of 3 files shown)


Notebooks/OpenIssues.github-issues
```diff
@@ -0,0 +1,42 @@
...
+  },
+  {
+    "kind": 1,
+    "language": "markdown",
+    "value": "All Open Issues with no assignee"
+  },
+  {
+    "kind": 2,
+    "language": "github-issues",
+    "value": "$repos is:open is:issue archived:false no:assignee"
```
makubacki, line 40: Could this one also track `state:needs-owner`?

So, no assignee or the needs owner label is present.

Notebooks/OpenIssues.github-issues
```diff
@@ -0,0 +1,42 @@
+[
```
makubacki, line 1: Can you also add a section for all stale issues?

`label:state:stale`

makubacki (approved):
Approved with some minor suggestions

Notebooks/OpenIssues.github-issues
apop5: Yes, that is a good idea.

Updated both PR and Issues to have stale queries.

Notebooks/OpenIssues.github-issues
apop5: I can make the query do that, but in testing, it doesn't give us what we want (a unified list of any issues that has either label:state:needs-owner and no:asignee)

Because label:state:needs-owner and no:asignee are separate query identifiers, it would create a query that only includes items that contain both.

If we could ensure that new issues get "label:state:needs-owner" automatically applied, we could just rely on that.

In the meantime, would having two separate queries satisfy our needs? Then we can remove the no:asignee query at a later date?

kuqin12 (approved):

Notebooks/OpenIssues.github-issues
makubacki: Separate queries is fine.

cfernald (approved):

PR #112, opened 2023-02-03
Re-introduce workaround to fix ADO path names for PR eval

cfernald:
Adds a work-around to the naming provided by System.PullRequest.targetBranch on ADO repos by stripping the refs/heads/ from the path since that makes it unrecognizable to git for the git diff used by stuart_pr_eval.

### DIFF
--- a/Steps/PrGate.yml
+++ b/Steps/PrGate.yml
@@ -87,12 +87,17 @@ steps:
 
 # Trim the package list if this is a PR
 - ${{ if eq(parameters.do_pr_eval, true) }}:
+  - powershell:
+      $TargetBranch = "$(System.PullRequest.targetBranch)".replace('refs/heads/', '');
+      Write-Host "##vso[task.setvariable variable=pr_compare_branch]origin/$TargetBranch";
+    displayName: Workaround for Branch Names
+    condition: eq(variables['Build.Reason'], 'PullRequest')
   - task: CmdLine@1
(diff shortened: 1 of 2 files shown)


Steps/PrGate.yml
```diff
@@ -87,12 +87,17 @@ steps:
 
 # Trim the package list if this is a PR
 - ${{ if eq(parameters.do_pr_eval, true) }}:
+  - powershell:
```
kuqin12, line 90: nit: :)

kuqin12 (approved):

cfernald:
I have been defeated by azure pipeline expressions. I attempted to do this inline in existing step, but could not get the combination of the replace expression with the pre-defined system variable to function properly. If anyone has any ideas on this please let me know.

os-d (approved):

Steps/PrGate.yml
cfernald: Sorry, not following what the nit is about here

makubacki (approved):

makubacki:
@cfernald what about the isntance in https://github.com/microsoft/mu_devops/blob/main/Steps/BuildPlatform.yml#L84?

cfernald:
> @cfernald what about the instance in https://github.com/microsoft/mu_devops/blob/main/Steps/BuildPlatform.yml#L84?

good point, will fix there as well

PR #115, opened 2023-02-04
.sync/azure_pipelines: Switch from microdnf to dnf

makubacki:
The following commit in tianocore/containers switched from `microdnf`
to `dnf` since `microdnf` is no longer available in the Fedora 35
minimal base image.

https://github.com/tianocore/containers/commit/3487a34

Therefore, `microdnf` will currently fail (not found). This change
updates code using the Fedora 35 container image to switch to `dnf`.


apop5 (approved):

makubacki:
Merging now to get to repos that need it and it is after hours

PR #118, opened 2023-02-06
.sync/workflows/leaf: Add PR formatting validator workflow

makubacki:
Adds a new workflow to validate pull request formatting.

### DIFF
--- a/.sync/Files.yml
+++ b/.sync/Files.yml
@@ -451,6 +451,28 @@ group:
     repos: |
       microsoft/mu_devops
 
+# Leaf Workflow - Pull Request Validator
+  - files:
+    - source: .sync/workflows/leaf/pull-request-formatting-validator.yml
+      dest: .github/workflows/pull-request-formatting-validator.yml
+    repos: |
+      microsoft/mu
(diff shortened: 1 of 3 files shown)


.sync/Files.yml
```diff
@@ -451,6 +451,28 @@ group:
     repos: |
       microsoft/mu_devops
 
+# Leaf Workflow - Pull Request Validator
+  - files:
+    - source: .sync/workflows/leaf/pull-request-formatting-validator.yml
+      dest: .github/workflows/pull-request-formatting-validator.yml
+    repos: |
+      microsoft/mu
```
apop5, line 459: Is there a way to make the repo list a variable, so that we can reuse the list?

apop5 (approved):

os-d (approved):

.sync/Files.yml
makubacki: Yes. I think that'd be a nice refactor. I filed a feature request here to track it, so it doesn't get lost.

https://github.com/microsoft/mu_devops/issues/121

PR #134, opened 2023-02-28
.sync: Add main branch as a trigger to CodeQL workflow

makubacki:
Some repos that use the workflow have a `main` branch instead of a
`release` branch as their primary branch.


makubacki:
Completing now since it's trivial and can get picked up in the current file sync.

Issue #84, opened 2023-01-13
[Bug] Code coverage results missing in pipeline UI


makubacki:
Some code coverage results are not reported in the Azure pipeline UI.

For example this build runs MacAddressEmulationDxe tests, but they do not appear in the coverage.xml 

- https://dev.azure.com/projectmu/mu/_build/results?buildId=34335&view=ms.vss-test-web.build-test-results-tab
Pipelines

There's also been changes to code coverage reporting and result file merge capabilities added to edk2 recently. See:

https://github.com/tianocore/edk2/pull/3349

This task tracks reviewing and updating code coverage for Project Mu repos.

makubacki:
@cfernald I recall the dotnet fix in the container image being related to this. Is it resolved now that #99 is merged? Or are additional changes needed?

cfernald:
@makubacki I think it depends slightly the original intent of this issue. This resolve the issue of the code coverage data no longer being collected, but there is still issues with the HTML UI not showing all of the code, so I think some more work is required in the HTML generation.

makubacki:
> there is still issues with the HTML UI not showing all of the code, so I think some more work is required in the HTML generation.

I agree that this issue should track resolving that as well.

cfernald:
Resolved with https://github.com/microsoft/mu_devops/pull/138

Issue #23, opened 2022-11-15
Test All Mu DevOps Consumer Builds in Pull Requests to Repo


makubacki:
We need to ensure that changes made in mu_devops pass all consumer pipelines.

Pull requests into mu_devops need to have status checks against all repos that use it as a repository resource to verify the build succeeds with the mu_devops change.

PR #138, opened 2023-03-08
Create code coverage post-processing steps and integrate into PR Jobs

cfernald:
1. Switch to coverage files per-matrix to not publish coverage
2. Add steps to download and merge/filter all artifacts
3. Integrates steps into job templates.

Issue #84

### DIFF
--- a/.sync/Version.njk
+++ b/.sync/Version.njk
@@ -35,5 +35,5 @@
 {# The latest Project Mu release branch value. #}
 {% set latest_mu_release_branch = "release/202208" %}
 
-{# The version of the fedora-35-build container to use. #}
+{# The version of the fedora-37-build container to use. #}
 {% set linux_build_container = "ghcr.io/tianocore/containers/fedora-37-build:f1c7a20" %}
--- a/.sync/Version.njk
+++ b/.sync/Version.njk
@@ -36,4 +36,4 @@
 {% set latest_mu_release_branch = "release/202208" %}
 
 {# The version of the fedora-35-build container to use. #}
-{% set linux_build_container = "ghcr.io/tianocore/containers/fedora-37-build:14d2aba" %}
+{% set linux_build_container = "ghcr.io/tianocore/containers/fedora-37-build:f1c7a20" %}
--- a/Jobs/PrGate.yml
+++ b/Jobs/PrGate.yml
@@ -96,6 +96,10 @@ jobs:
       - script: echo "##vso[task.prependpath]/home/vsts_azpcontainer/.local/bin"
         displayName: Add User Local Bin to Path
     - ${{ parameters.extra_steps }}
+    - script: |
+              sudo dnf install --assumeyes lcov
+              pip install lcov_cobertura
+      displayName: Install Code Coverage Tools
     - template: ../Steps/PrGate.yml
       parameters:
         artifacts_identifier: '$(package) $(target)'
@@ -110,3 +114,15 @@ jobs:
         do_pr_eval: ${{ parameters.do_pr_eval }}
         tool_chain_tag: ${{ parameters.tool_chain_tag }}
         install_tools: ${{ not(parameters.container_build) }}
+        extra_build_args: CODE_COVERAGE=TRUE
+
+  - job: PublishCoverage
+    dependsOn: Build
+    workspace:
+      clean: all
+
+    pool:
+      vmImage: 'windows-2022'
+
+    steps:
+      - template: ../Steps/PublishCodeCoverage.yml
--- a/Jobs/PrGate.yml
+++ b/Jobs/PrGate.yml
@@ -47,7 +47,7 @@ parameters:
 - name: linux_container_image
   displayName: Linux Container Image
   type: string
-  default: 'ghcr.io/tianocore/containers/fedora-37-build:f1c7a20'
+  default: ''
 - name: packages
   displayName: Packages
   type: string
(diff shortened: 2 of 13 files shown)


cfernald:
Opening as draft for reference, will likely be updated for tooling changes.

kuqin12 (approved):

apop5 (approved):

.sync/Version.njk
```diff
@@ -36,4 +36,4 @@
 {% set latest_mu_release_branch = "release/202208" %}
 
 {# The version of the fedora-35-build container to use. #}
-{% set linux_build_container = "ghcr.io/tianocore/containers/fedora-37-build:14d2aba" %}
+{% set linux_build_container = "ghcr.io/tianocore/containers/fedora-37-build:f1c7a20" %}
```
makubacki, line 39: Please hold until Monday for me to look then.

Jobs/PrGate.yml
```diff
@@ -46,7 +47,7 @@ parameters:
 - name: linux_container_image
   displayName: Linux Container Image
   type: string
-  default: 'ghcr.io/tianocore/containers/fedora-35-build:5b8a008'
+  default: 'ghcr.io/tianocore/containers/fedora-37-build:f1c7a20'
```
makubacki, line 50: I ultimately want to get to the point where we track the exact container image (by SHA) in one place. Right now that is Version.njk.

My thought was:

1. Any pipeline that cares to use a specific version should pass it in as a parameter (Version.njk is one way to get the value and do that)
2. The default requires less maintenance over time

With that in mind, can this just be `"fedora-37-build:latest"` for now? As long as latest works as expected in the tianocore container repo (it does not always depending on how you manage images).

Jobs/PrGate.yml
```diff
@@ -110,3 +111,15 @@ jobs:
...
         install_tools: ${{ not(parameters.container_build) }}
+        extra_build_args: CODE_COVERAGE=TRUE
+
+  - job: PublishCoverage
+    dependsOn: Build
+    workspace:
+      clean: all
+
+    pool:
+      vmImage: 'windows-2022'
```
makubacki, line 122: If this doesn't have VS toolchain dependencies (it does not) then I suggest to use "`windows-latest`". It will be one less place to fall behind and cause problems in the future when "`windows-2022`" has brown outs and is eventually deprecated.

Steps/PublishCodeCoverage.yml
```diff
@@ -0,0 +1,45 @@
...
+
+parameters:
+- name: include_paths
+  displayName: Tool Chain (e.g. VS2022)
+  type: string
+  default: ''
+- name: exclude_paths
+  displayName: Tool Chain (e.g. VS2022)
+  type: string
+  default: ''
```
makubacki, line 16: `include_paths` and `exclude_paths` are not used in this file. Also the `displayName` comment seems unrelated.

cfernald: Yes, I decided to drop this in one of the stages and forgot to remove the parameters

Jobs/PrGate.yml
cfernald: My biggest concern with the latest is that if we support forked branches, they would keep getting the new container but with all the old build definitions. It's possible there could be incompatibilities there we would have to concern ourselves with like dropping tools that are no longer used, etc.

makubacki: Does any code depend on the default value right now? Ultimately, I think all template users should pass this value since they need to account for changes in the image. For example, this could be an empty default and we raise a pipeline error if it remains empty.

If the repo opts in to receive its "leaf" pipelines from file sync, they get it automatically (from Version.njk). Otherwise, they can pull Version.njk down in their pipeline run and extract the value.

makubacki: > Is pulling version.njk feasible for Azure dev-ops repos?

Sure. Here's a quick curl example:
```bash
curl -s https://raw.githubusercontent.com/microsoft/mu_devops/main/.sync/Version.njk | grep linux_build_container | cut -d '"' -f 2
```

What I'm proposing is that we do this one way. Either (1) require template users to specify what container image they want or (2) always specify the container image in the pipeline.

By this being a template parameter, that means (1) has been chosen. So, the complexity associated with testing that is inherited. (2) brings more complexity that I don't think is needed if (1) is supported. I think (1) makes sense because it allows individual repos to control their build environment.

Then, it becomes a testing problem. Which is an issue (#23). But, because this repo produces versioned release, dependent repos do not need to update what mu_devops they use until they can verify their chosen container image works with the mu_devops version they are updating to and that can be confirmed in their CI build process.

cfernald: Yeah, I suppose I see this as a convenience. For all the synced repos it's moot really since they get their top-level pipeline files from this repo anyways so the burden of testing really doesn't change one way or the other in my opinion. For all the non-synced repos this allows an easy way to just stay in lock-step with dev_ops. All such repos will need to manually roll their dev_ops version anyways, so whether they need to validate the container separately or not seems to make no difference testing-wise in my mind. This just provides an easy solution if they don't need container customizations.

Either way, I don't think there are any non-synced repros that use containers anyways, so i'm happy to just push this to the azure_pipelines/custom/* files using version.njk if that work for you.

Steps/InstallCoverageTools.yml
```diff
@@ -0,0 +1,14 @@
+## @file
```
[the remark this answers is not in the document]
cfernald: Agreed, this type of documentation I think will live in the basecore/mu documentation though.

makubacki (approved):

PR #143, opened 2023-03-13
.sync/Files.yml: Move CodeQL workflow comment

makubacki:
Move out of the `repos` list and append onto the preceding group
description.


makubacki:
Merging now to fix workflow

PR #154, opened 2023-04-06
Adding support for self-hosted pools

kuqin12:
This change added support for self-hosted agent pools for potentially onboarding Windows on ARM systems as there is no plan for Microsoft hosted Windows ARM agents as of today.


Jobs/PrGate.yml
```diff
@@ -85,7 +89,7 @@ jobs:
       clean: all
 
     pool:
-      vmImage: ${{ parameters.vm_image }}
```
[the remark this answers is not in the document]
kuqin12: Updated.

.sync/Files.yml
```diff
@@ -27,83 +27,35 @@
...
-      template: true
-    - source: .sync/azure_pipelines/custom/mu_oem_sample/Windows-VS.yml
-      dest: .azurepipelines/Windows-VS.yml
-      template: true
-    repos: |
-      microsoft/mu_oem_sample
-
-# Azure Pipelines - Matrix Dependent - All
+# Azure Pipelines - Matrix Dependent - CLANDPDB
   - files:
```
cfernald, line 31: Should we convert these files as well?

kuqin12: Or should we remove them from this repo (and not do the sync)? We do not seem to run them anyway?

cfernald: Or delete them if they are not used. @makubacki , it looks like these were converted but I don't see any pipeline. was that intentional?

Jobs/PrGate.yml
```diff
@@ -68,64 +64,94 @@ parameters:
...
+        name: $(${{ parameters.pool_name }})
+        demands:
+        - Agent.OS -equals ${{ parameters.os_type }}
+    ${{ else }}:
+      pool:
+        vmImage: ${{ parameters.vm_image }}
 
     # Use a container if one was specified.
-    ${{ if and(eq(parameters.container_build, true), not(contains(parameters.vm_image, 'windows'))) }}:
+    ${{ if and(eq(parameters.container_build, true), not(contains(parameters.vm_image, 'windows')), not(eq(item.Value.SelfHostAgent, true))) }}:
```
cfernald, line 117: nit: consider using ne() instead of not(eq())

cfernald: below as well

Jobs/PrGate.yml
```diff
@@ -68,64 +64,94 @@ parameters:
...
+- name: build_matrix
+  type: object
+  # Each element in the matrix will be used in a separate job
+  # Required fields:
+  # Pkgs:           Specify what packages in the target repo are to be build for this job
+  # Targets:        Specify which targets are to be built in this job, i.e. DEBUG,
+  #                 RELEASE, NO-TARGET, NOOPT. The acceptable values depend on the
+  #                 Pkgs to be built.
+  # ArchList:       List of architectures to be supported by the packages (e.g. IA32, X64)
+  # SelfHostAgent:  A boolean indicating whether this job should be run on the selfhosted
```
cfernald, line 91: Is it possible to make SelfHostAgent conditional? If this value is not prevent then eq(item.Value.SelfHostAgent, true)) would be false still right? or just it fail?

kuqin12: Agreed, this is updated.

Jobs/PrGate.yml
kuqin12: Agreed, updated.

.sync/azure_pipelines/MuDevOpsSync.yml
```diff
@@ -0,0 +1,90 @@
...
+# - Mu DevOps Repo: https://github.com/microsoft/mu_devops
+# - File Sync Settings: https://github.com/microsoft/mu_devops/blob/main/.sync/Files.yml
+#
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+{% import '../../../Version.njk' as sync_version -%}
+
+resources:
```
cfernald, line 16: @makubacki , would be curious to get your input on this. This file exists pretty much to get the devops version. Kun explored a number of ways to do this without having to sync a wrapper template, but it doesnt seem practical. Any thoughts on how this could be done?

cfernald (approved):

.sync/Files.yml
makubacki: They were added to bring existing support into mu_devops that was in mu_basecore. I agree that they're not used at the moment, and I'd suggest to delete them as well.

makubacki:
I will look at this in detail on Tuesday.

Steps/PrGate.yml
```diff
@@ -112,6 +116,10 @@ steps:
 - ${{ if eq(parameters.install_tools, true) }}:
   - template: InstallCoverageTools.yml
 
+# Potential Extra steps
+- ${{ if eq(parameters.self_host_agent, false) }}:
```
makubacki, line 120: Is it a general rule that self-hosted agents do not potentially need extra install steps?

Steps/PrGate.yml
```diff
@@ -133,9 +141,6 @@ steps:
     script: stuart_update -c ${{ parameters.build_file }} -p $(pkgs_to_build) -t ${{ parameters.build_targets}} -a ${{ parameters.build_archs}} TOOL_CHAIN_TAG=${{ parameters.tool_chain_tag}}
   condition: and(gt(variables.pkg_count, 0), succeeded())
 
-# Potential Extra steps
```
makubacki, line 136: There's nothing specifically bad about moving this, but was it fixing some existing use case or enabling a new one?

This alone could make this change a breaking change in my opinion, if, for example, the extra step assumed Stuart `setup` and `update` had already run.

Jobs/PrGate.yml
```diff
@@ -68,64 +64,99 @@ parameters:
...
+  type: string
+  values:
+  - Windows_NT
+  - Linux
+  default: Windows_NT
+- name: build_matrix
+  type: object
+  # Each element in the matrix will be used in a separate job
+  # Required fields:
+  # Pkgs:           Specify what packages in the target repo are to be build for this job
```
makubacki, line 86:
```suggestion
  # Pkgs:           Specify what packages in the target repo are to be built for this job
``

File: .sync/Files.yml [hunk 9]
```diff
@@ -27,83 +27,35 @@
...
-      dest: .azurepipelines/Ubuntu-GCC5.yml
-      template: true
-    - source: .sync/azure_pipelines/custom/mu_oem_sample/Windows-VS.yml
-      dest: .azurepipelines/Windows-VS.yml
-      template: true
-    repos: |
-      microsoft/mu_oem_sample
-
-# Azure Pipelines - Matrix Dependent - All
+# Azure Pipelines - Matrix Dependent - CLANDPDB
```
makubacki, line 30: We discussed dropping the `CLANGPDB` files. In case they are not.


```suggestion
# Azure Pipelines - Matrix Dependent - CLANGPDB
```

.sync/azure_pipelines/MuDevOpsSync.yml
makubacki: If I understand correctly, I don't think so.

1. The `resources` section is processed very early to set up the pipeline and template binding.
2. We decided to fix repos on an exact version of mu_devops so that needs to be recorded in the repo's history.

I believe (1) eliminates most dynamic options and (2) means a sync of the mu_devops version to the repo (of some kind) is needed.

As the scripts are currently written, I think this route is acceptable to reduce number of files synced and still control the version in one location (`Version.njk`).

.sync/azure_pipelines/MuDevOpsSync.yml
```diff
@@ -0,0 +1,90 @@
+## @file
+# Azure Pipeline build file for a build using mu_devops.
```
makubacki, line 2: I don't know if the fact that the file is synced from `mu_devops` should be its defining characteristic (its filename).

Maybe something like `MuDevOpsWrapper.yml` would be a little more precise when looking at the file on its own in a given repo.

Jobs/PrGate.yml
```diff
@@ -68,64 +64,99 @@ parameters:
...
+  type: string
+  default: ''
+- name: os_type
+  displayName: OS type on the self-hosted agent pools
+  type: string
+  values:
+  - Windows_NT
+  - Linux
+  default: Windows_NT
+- name: build_matrix
```
makubacki, line 82: I think this comment (or somewhere in this repo's documentation) should give a brief example of a matrix.

.sync/Files.yml
kuqin12: Thanks, will remove them soon.

Steps/PrGate.yml
kuqin12: That is the current approach. Until when we switch to a full container approach, I think self-hosted agents should meet the prerequisites of tooling. (other than pip modules as they are too dynamic).

Steps/PrGate.yml
kuqin12: This is to resolve some steps needed to be run before the stuart_setup/stuart_update. For example, if the consumer would like to authenticate a nuget feed to complete the update step, the extra installation steps will need to be inserted here.

Generally, I think the extra_install_steps should happen after the generic install steps, which is where this change is located. But I agree that this is at risk of breaking existing pipelines at large. A layman question next: how do we track breaking changes/make that more obvious in this repo?

.sync/Files.yml
kuqin12: Will remove shortly

.sync/azure_pipelines/MuDevOpsSync.yml
kuqin12: thanks, will udpate!

Jobs/PrGate.yml
kuqin12: Agreed, will update.

.sync/Files.yml
kuqin12: The CLANGPDB files are gone.

.sync/Files.yml
kuqin12: These steps are removed.

.sync/azure_pipelines/MuDevOpsSync.yml
kuqin12: The file name is updated.

Steps/PrGate.yml
makubacki: If that's the case, what do you think about raising an error if the caller has `parameters.self_host_agent` set to `true` and `parameters.extra_install_step` is non-empty?

Steps/PrGate.yml
makubacki: If you're not seeing issues in testing we should be good then and mark is a breaking change. You can manually add the `impact:breaking-change` label.

makubacki (approved):

Steps/PrGate.yml
kuqin12: The `parameters.extra_install_step` is not set per matrix entry after the change, but instead being used for each `jobs`. But the `parameters.self_host_agent` is tied to each matrix entry.

This was done because I think it is undesirable to duplicate the pipelines just for selfhost builds, because those will be manual process and hard to manage. Expanding the matrix would be an easier path for repos to enroll if they want.

On the other hand, because the wrapper yaml file (the one that will be synced) contains `resources`, we can only use `extends` instead of `jobs -template` to apply it. This limited the capability of using that file as a template twice in one place.

Steps/PrGate.yml
kuqin12: Thanks, I added the breaking-change tag and will update the other repos then test this again.

Jobs/PrGate.yml
kuqin12: Added two examples of matrix entries in the comment, one that uses self host agent, one that does not.

.sync/Files.yml
```diff
@@ -27,83 +27,24 @@
...
-      microsoft/mu_basecore
-
 # Azure Pipelines - Matrix Dependent - GCC & VS Only
   - files:
-    - source: .sync/azure_pipelines/matrix_dependent/Ubuntu-GCC5.yml
-      dest: .azurepipelines/Ubuntu-GCC5.yml
-      template: true
-    - source: .sync/azure_pipelines/matrix_dependent/Windows-VS.yml
-      dest: .azurepipelines/Windows-VS.yml
+    - source: .sync/azure_pipelines/MuDevOpsSync.yml
```
kuqin12, line 32: Needs update file name here.

kuqin12: Updated.

Steps/PrGate.yml
kuqin12: Synced offline on this and added warning message to the `Stesp/PrGate.yml` to make skipped steps more obvious for easier troubleshooting.

Steps/PrGate.yml
kuqin12: All consumer repos are updated and testing in progress. For the status of consuming PRs, please check the links in https://github.com/microsoft/mu_devops/pull/154#issuecomment-1560077672.

PR #167, opened 2023-04-27
Add quotes to git safe.directory in .devcontainer to fix parsing error

cfernald:
Currently the git command fails because the * is not parsed correctly in command line without the quote. This resolves this issue.

### DIFF
--- a/.sync/devcontainer/devcontainer.json
+++ b/.sync/devcontainer/devcontainer.json
@@ -1,6 +1,6 @@
 {
   "image": "ghcr.io/tianocore/containers/fedora-37-dev:latest",
-  "postCreateCommand": "git config --global --add safe.directory * && pip install --upgrade -r pip-requirements.txt",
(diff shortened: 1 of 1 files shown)


apop5 (approved):

.sync/devcontainer/devcontainer.json
```diff
@@ -1,6 +1,6 @@
 {
   "image": "ghcr.io/tianocore/containers/fedora-37-dev:latest",
-  "postCreateCommand": "git config --global --add safe.directory * && pip install --upgrade -r pip-requirements.txt",
+  "postCreateCommand": "git config --global --add safe.directory '*' && pip install --upgrade -r pip-requirements.txt",
```
makubacki, line 3: According to [this SO post](https://stackoverflow.com/a/71904131), double quotes might be needed on Windows. Did you see an issue like that?

makubacki: We can go with that for now.

makubacki (approved):

PR #168, opened 2023-04-27
.sync/codeql: Attempt to upload logs in case of success and failures

makubacki:
Currently, log artifacts are only uploaded in case of success. This
change also uploads logs in case of failures to aid in debugging.

### DIFF
--- a/.sync/workflows/leaf/codeql.yml
+++ b/.sync/workflows/leaf/codeql.yml
@@ -153,12 +153,42 @@ jobs:
       if: steps.get_ci_file_operations.outputs.setup_supported == 'true'
       run: stuart_setup -c .pytool/CISettings.py -t DEBUG -a ${{ matrix.archs }} TOOL_CHAIN_TAG=${{ matrix.tool_chain_tag }}
 
+    - name: Upload Setup Log As An Artifact
+      uses: actions/upload-artifact@v3
+      if: (success() || failure()) && steps.get_ci_file_operations.outputs.setup_supported == 'true'
+      with:
+        name: ${{ matrix.package }}-Logs
+        path: |
+          **/SETUPLOG.txt
+          retention-days: 7
+        if-no-files-found: ignore
+
     - name: CI Setup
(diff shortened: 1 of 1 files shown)


kuqin12 (approved):

.sync/workflows/leaf/codeql.yml
```diff
@@ -153,13 +153,43 @@ jobs:
...
+          retention-days: 7
+        if-no-files-found: ignore
+
     - name: CI Setup
       if: steps.get_ci_file_operations.outputs.ci_setup_supported == 'true'
       run: stuart_ci_setup -c .pytool/CISettings.py -t DEBUG -a ${{ matrix.archs }} TOOL_CHAIN_TAG=${{ matrix.tool_chain_tag }}
 
+    - name: Upload CI Setup Log As An Artifact
+      uses: actions/upload-artifact@v3
+      if: (success() || failure()) && steps.get_ci_file_operations.outputs.ci_setup_supported == 'true'
```
os-d, line 172: Why do some steps depend on `steps.get_ci_file_operations.outputs.ci_setup_supported == 'true'` and not others?

os-d (approved):

apop5 (approved):

.sync/workflows/leaf/codeql.yml
makubacki: A Python script in an earlier step outputs `ci_setup_supported` by parsing the CISettings.py file in the repo. Not all CISettings.py files subclass `CiSetupSettingsManager` and support CI setup commands. For example, [mu_feature_dfci](https://github.com/microsoft/mu_feature_dfci/blob/main/.pytool/CISettings.py) supports it and [mu_basecore](https://github.com/microsoft/mu_basecore/blob/release/202208/.pytool/CISettings.py) does not.

The two steps that have this condition are `CI Setup` and `Upload CI Setup Log As An Artifact`. The first won't work if not supported, the second won't find any logs if the first doesn't run (not fatal though because it specifies `if-no-files-found: ignore`).

PR #171, opened 2023-04-28
<Binary/Other>CopyAndPublish.yml: Use bash to parse parameter string

makubacki:
Uses bash to parse a parameter string that could be empty since it
is more robust and consistent than directly depending on pipeline
string interpretation logic.

This treats the following strings as "empty":
- ""
- " "


makubacki:
Note: Tested in PR https://github.com/microsoft/mu_tiano_platforms/pull/525

kenlautner (approved):

TaylorBeebe (approved):

PR #175, opened 2023-05-17
Initial check-in for MU Linux container and pipelines

cfernald:
Creates a Ubuntu-22 based image for building Project MU. These changes are largely based on the infrastructure from [tianocore/containers](https://github.com/tianocore/containers).

Sizes:
Build -  2.74 Gb
Test - 3.63 Gb
Dev - 3.63 Gb


apop5 (approved):

Containers/Readme.md
makubacki, line 1: @cfernald please hold the PR for about a day or so for me to come back and review it.

makubacki:
@cfernald was there a particular reason for choosing Ubuntu?

I think it's fine, but it would be helpful to describe in the containers readme file why that distro was chosen for Mu Linux builds.

.github/workflows/Container-Ubuntu-22.yml
```diff
@@ -0,0 +1,28 @@
...
+name: "Ubuntu 22 Containers"
+
+# This workflow only runs (on the main branch or on PRs targeted
+# at the main branch) and if files inside the Ubuntu-22 directory
+# have been modifed/added/removed...
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [ main ]
```
makubacki, line 15: If you're going to use a multi-line array for `paths`, it'd be consistent to do the same for `branches`.

```suggestion
    branches:
      - main

```

.github/workflows/Container-Ubuntu-22.yml
```diff
@@ -0,0 +1,28 @@
...
+    paths:
+      - 'Containers/Ubuntu-22/**'
+  pull_request:
```
makubacki, line 18: Are you sure that this needs a `pull_request` trigger?

Since the `pull_request` trigger runs on open, synchronize, and reopened types by default and this is a public repo that could lead to a lot of images potentially being created that are not useful.

.github/workflows/Container-Ubuntu-22.yml
makubacki, line 1: Since we're eventually going to need to support multiple containers, we can probably reduce overhead if this file could support multiple containers that might need to be built.

For example, if I run `gh pr diff 175 --name-only` (for this PR) it returns:

```
> gh pr diff 175 --name-only
.github/workflows/Container-Ubuntu-22.yml
.github/workflows/container-build.yml
Containers/Readme.md
Containers/Ubuntu-22/Dockerfile
```

Whatever the context (PR, commit, etc.), we could get a list of files modified. Look for `Containers/<container_name>` and then use the list of `<container_name>`s found to pass into `container-build.yml`.

cfernald:
> @cfernald was there a particular reason for choosing Ubuntu?
> 
> I think it's fine, but it would be helpful to describe in the containers readme file why that distro was chosen for Mu Linux builds.

2 reason, first Fedora has limited support cross-compiling usermode components which is problematic for basetools (which may not matter), but also people are just generally more familiar and aligns with peoples existing development flows better.

makubacki:
> > @cfernald was there a particular reason for choosing Ubuntu?
> > I think it's fine, but it would be helpful to describe in the containers readme file why that distro was chosen for Mu Linux builds.
> 
> 2 reason, first Fedora has limited support cross-compiling usermode components which is problematic for basetools (which may not matter), but also people are just generally more familiar and aligns with peoples existing development flows better.

Please add this info to the containers readme file.

.github/workflows/Container-Ubuntu-22.yml
cfernald: Thats an interesting idea. I suppose it would mean merging all the pipelines together and using a matrix, but i'm not really opposed to that.

.github/workflows/Container-Ubuntu-22.yml
makubacki: I see the `if: ${{ github.ref == 'refs/heads/main' }}` push condition now. That's fine.

PR #188, opened 2023-06-01
Fixing the sync file being empty and update the version to be used

kuqin12:
This change fixed an issue on auto-populated sync files by adding raw text tags around the applicable sections.

It also updated the tag version to v4.0.0 to pick up the latest change of supporting self-hosted agents.

### DIFF
--- a/.sync/Version.njk
+++ b/.sync/Version.njk
@@ -30,7 +30,7 @@
 #}
 
 {# The git ref value that files dependent on this repo will use. #}
-{% set mu_devops = "v3.0.0" %}
+{% set mu_devops = "v4.0.0" %}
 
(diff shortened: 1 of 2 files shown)


kuqin12:
Tested this file locally and the output was generated as expected now.

cfernald (approved):

makubacki (approved):

makubacki:
The raw text is an easy one to miss given the two syntaxes commonly used in these files.

cfernald:
> The raw text is an easy one to miss given the two syntaxes commonly used in these files.

I wonder if there is a way to make nunjucks blow up if it finds a variable it can't find to avoid this kind of thing in the future happening by accident.

PR #198, opened 2023-06-07
Publish coverage on even with skipped builds

kuqin12:
Current pipeline job for publishing coverage will skip the jobs if there are a few jobs being skipped. However, some skips are expected, such as some jobs that should be run on selfhosted agents, but no agents are available on the public site.

This change will update the publish coverage job's condition to ignore the skipped conditions, such that it will always run the job as long as there the previous job(s) not failing or being cancelled.

### DIFF
--- a/Jobs/PrGate.yml
+++ b/Jobs/PrGate.yml
@@ -163,7 +163,7 @@ jobs:
   dependsOn:
   - ${{ each item in parameters.build_matrix }}:
     - Build_${{ item.Key }}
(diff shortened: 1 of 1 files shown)


kuqin12:
Example runs here:
Succeeded: https://dev.azure.com/projectmu/mu/_build/results?buildId=48570&view=results
Cancelled: https://dev.azure.com/projectmu/mu/_build/results?buildId=48575&view=results

cfernald (approved):

makubacki (approved):

PR #220, opened 2023-07-14
Added fetchdepth to disable default shallow checkout in new pipelines

apop5:
https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-checkout?view=azure-pipelines#shallow-fetch

`New pipelines created after the [September 2022 Azure DevOps sprint 209 update](https://learn.microsoft.com/en-us/azure/devops/release-notes/2022/sprint-209-update) have Shallow fetch enabled by default and configured with a depth of 1. Previously the default was not to shallow fetch. To check your pipeline, view the Shallow fetch setting in the [pipeline settings UI](https://learn.microsoft.com/en-us/azure/devops/pipelines/repos/azure-repos-git#shallow-fetch).`

Adding fetchdepth of 0 to disable default of shallow checkout for new pipelines using mu_devops.

### DIFF
--- a/Jobs/CreateBuildMatrix.yml
+++ b/Jobs/CreateBuildMatrix.yml
@@ -36,6 +36,7 @@ jobs:
 
     steps:
     - checkout: none
+       fetchDepth: 0
 
     - task: PowerShell@2
       name: CalculateMatrix
--- a/Jobs/CreateBuildMatrix.yml
+++ b/Jobs/CreateBuildMatrix.yml
@@ -36,7 +36,7 @@ jobs:
 
     steps:
     - checkout: none
-       fetchDepth: 0
+      fetchDepth: 0
 
     - task: PowerShell@2
       name: CalculateMatrix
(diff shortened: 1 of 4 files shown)


Javagedes (approved):

Jobs/CreateBuildMatrix.yml
makubacki, line 1: We were previously disabling shallow fetch in the pipeline UI for most pipelines I'm aware of.

Typically, we do want to disable shallow fetch but this setting overrides the pipeline UI so it seems a bit less flexible than just letting it be controlled per pipeline.

This is just being made as a matter of convenience, right?

makubacki (approved):

Jobs/CreateBuildMatrix.yml
makubacki: For this set of pipelines, I'm fine to move forward with the change.

apop5: In addition to being a matter of convenience, this is also trying to ease new pipeline creation and to proof against future pipeline configuration modifications.

This change is only trying to target the specific checkout steps that need a full checkout. The helper ymls that only require shallow checkouts still have a fetchdetch that they set.

PR #222, opened 2023-07-25
Create simple tag generator for ADO pipelines

cfernald:
Creates a script and pipeline that allows for automated tagging based on release version and commit messages. This template should be called for an Azure Devops hosted repository. The consumer of this template is expected to ensure the proper permissions for the build agent to be able to create the tag and commit tag notes.

The TagGenerator Script provides the following functionality:
- Scans git history for the most recent matching tag, e.g. 202302.5.10
- Generates the new tag version, incrementing the "major" version based on the presence of breaking changes.
- Generates release notes including commits by type, links to Azure Devops PRs, and contributors.


Jobs/GenerateTag.yml
```diff
@@ -0,0 +1,51 @@
...
+        clean: true
+        fetchTags: true
+        persistCredentials: true
+        path: "target"
+
+      - checkout: mu_devops
+        path: "mu_devops"
+        fetchDepth: 0
+
+      - task: UsePythonVersion@0
```
apop5, line 27: Can UsePythonVersion get replaced with [Steps](https://github.com/microsoft/mu_devops/tree/main/Steps)
/SetupPythonPreReqs.yml?

- template: Steps/SetupPythonPreReqs.yml
  install_pip_modules: false

cfernald:
This is published as a proof of concept. It will get much more work before being published.

cfernald:
Example of the release notes:

image

Jobs/GenerateTag.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - script: |
+          python mu_devops/Scripts/TagGenerator.py -r target/ -b ${{ parameters.base_version }} -v --tagvar tag_name --notes target/${{ parameters.notes_file }} --url $(Build.Repository.Uri)
+        displayName: "Run Tag Generator"
+        workingDirectory: $(Agent.BuildDirectory)
+
+      - script: |
+          set -e
+          git branch
+          git add ReleaseNotes.md
```
cfernald, line 62: oops, should use parameter

Scripts/TagGenerator.py
```diff
@@ -0,0 +1,211 @@
...
+            VerboseLog(
+                f"Different base version. {version_split[0]} -> {args.base}")
+
+    version = f"{args.base}.{major}.{minor}"
+    Log(f"New tag: {version}")
+
+    # Before going further, ensure this is not a duplicate. This can happen if
+    # there are tags detached from their intended branch.
+    for tag in repo.tags:
+        if f"{tag}" == version:
```
cfernald, line 70: Should use tag.name

Jobs/GenerateTag.yml
```diff
@@ -0,0 +1,69 @@
+## @file
+# Template file used to generate tags on ADO. This template requires that the
+# consumer specifies this repository as a resource named mu_devops.
+#
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
kuqin12, line 5: nit: I thought we do not do "All rights reserved." here?

Jobs/GenerateTag.yml
```diff
@@ -0,0 +1,69 @@
...
+    type: string
+    default: "ReleaseNotes.md"
+
+jobs:
+  - job: Create_Release_Tag
+    steps:
+      - checkout: self
+        clean: true
+        fetchTags: true
+        persistCredentials: true
```
kuqin12, line 33: Just curious, what is this for?

Jobs/GenerateTag.yml
```diff
@@ -0,0 +1,69 @@
...
+          pip install GitPython
+        displayName: "Install Dependencies"
+
+      - script: |
+          git config --global user.name "${{ parameters.git_name }}"
+          git config --global user.email "${{ parameters.git_email }}"
+        displayName: "Setup Git"
+
+      - script: |
+          python mu_devops/Scripts/TagGenerator.py -r target/ -b ${{ parameters.base_version }} -v --tagvar tag_name --notes target/${{ parameters.notes_file }} --url $(Build.Repository.Uri)
```
kuqin12, line 55: Should this be `$(tag_name)`?

Jobs/GenerateTag.yml
cfernald: This is required in order to push the generated tag and notes commit.

Jobs/GenerateTag.yml
cfernald: No, perhaps a bit hacky but I am actually passing the name of the variable for the python script to set it. A bit odd perhaps but I think its cleaner then trying to parse it out from the script output personally.

kuqin12: ah, name of the variable :) that is good to know.

Jobs/GenerateTag.yml
```diff
@@ -0,0 +1,69 @@
...
+        path: "target"
+
+      - checkout: mu_devops
+        path: "mu_devops"
+        fetchDepth: 0
+
+      - template: ../Steps/SetupPythonPreReqs.yml
+        parameters:
+          install_pip_modules: false
+
```
MarcChen46, line 43: maybe we should add one step here to validate the base_version, git_name and git_email parameters since consumer  should update these parameters instead of leave it as default empty string.

Scripts/TagGenerator.py
```diff
@@ -0,0 +1,211 @@
...
+from git import Repo
+from datetime import datetime
+import time
+
+parser = argparse.ArgumentParser()
+
+parser.add_argument("-r", "--repo", default=".",
+                    help="Path to the repo directory.")
+parser.add_argument("-b", "--base", type=str, required=True,
+                    help="The base release version. This must be provided")
```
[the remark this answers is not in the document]
cfernald: The original intention was to keep major for the "breaking", but without the need to concatinate the EKII release tag version with the breaking changes version. It isn't really as clear as I'd hoped though so I think you are right. I will change to major.minor.patch, but semantically the minor will be breaking changes. 

Alternatively we could conform to the versioning standard we use on the git repros, but I fear this might add some confusion on the consumers here since the distinction between features and bug fixes may not be useful.

Scripts/TagGenerator.py
```diff
@@ -0,0 +1,211 @@
...
+parser.add_argument("-n", "--notes", type=str,
+                    help="Provides path to the release notes markdown file.")
+parser.add_argument("--tagvar", type=str,
```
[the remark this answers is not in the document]
cfernald: Will do

Jobs/GenerateTag.yml
```diff
@@ -0,0 +1,76 @@
...
+    steps:
+      - checkout: self
+        clean: true
+        fetchTags: true
+        persistCredentials: true
+        path: "target"
+
+      - checkout: mu_devops
+        path: "mu_devops"
+        fetchDepth: 0
```
apop5, line 43: I think fetchepth can be omitted or set to 1 to reduce the amount of data done in the checkout. We only need the latest mu_devops.

apop5 (approved):

makubacki (approved):

PR #247, opened 2023-09-21
RustSetupSteps: Add target triple detection

makubacki:
Adds a step to determine the target triple for the host platform
that can be reused by other steps. It is currently used in the
`rustup component add` command issued later in the file.

### DIFF
--- a/.sync/azure_pipelines/RustSetupSteps.yml
+++ b/.sync/azure_pipelines/RustSetupSteps.yml
@@ -28,6 +28,7 @@ steps:
     arch = platform.machine()
 
     rust_targets = {
+        ('Windows', 'aarch64'): 'aarch64-pc-windows-msvc',
         ('Windows', 'x86_64'): 'x86_64-pc-windows-msvc',
         ('Windows', 'i386'): 'i686-pc-windows-msvc',
         ('Linux', 'x86_64'): 'x86_64-unknown-linux-gnu',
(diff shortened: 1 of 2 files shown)


.sync/azure_pipelines/RustSetupSteps.yml
```diff
@@ -17,6 +17,30 @@
...
+  run: |
+    import os
+    import platform
+
+    system = platform.system()
+    arch = platform.machine()
+
+    rust_targets = {
+        ('Windows', 'x86_64'): 'x86_64-pc-windows-msvc',
+        ('Windows', 'i386'): 'i686-pc-windows-msvc',
```
kuqin12, line 32: we will have aarch64 msvc as well?

kuqin12 (approved):

apop5 (approved):

.sync/azure_pipelines/RustSetupSteps.yml
makubacki: It's possible. I was looking at [tier 1 targets](https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-1-with-host-tools) which doesn't have `aarch64-pc-windows-msvc` (yet).

It is a tier target though so I suppose I could add it now to have it covered. I'll make that update.

makubacki: Updated

makubacki:
Fyi, pipelines are backed up at the moment. I need to get results from another pending run before completing this.

Javagedes (approved):

PR #258, opened 2023-09-27
Add UploadCodeCoverage Template 

### LINKED ISSUE #257: Add Rust coverage data to PRs
We can generate coverage data today with cargo tarpaulin. This task tracks adding that coverage data to codecov.io so we can get PR comments showing coverage delta.

This item is meant to track the overall task, though the changes may not be entirely in mu_devops.

Javagedes:
Add a new template, UploadCodeCoverage, that can be used to upload a code coverage document (such as cobertura) to codecov.io.

Uses the `-Z` flag, which will cause the step to fail if uploading fails for any reason.

See: https://dev.azure.com/projectmu/mu/_build/results?buildId=55185&view=results for a working example.

Closed #257 but this should be usable in non-rust environments. I will start uploading all code coverage (such as for mu_basecore) once I have `stuart_report` merged and being used to clean up the code coverage data.


Steps/UploadCodeCoverage.yml
```diff
@@ -0,0 +1,31 @@
...
+  displayName: Codecov.io Token
+  type: string
+  default: ''
+- name: report_dir
+  displayName: Code Coverage Report
+  type: string
+  default: ''
+
+steps:
+- powershell: |
```
makubacki, line 19: While we're trying to reduce dependencies on Powershell overall, I'm fine with this since its usage is conditionalized to Windows. Another option would be to write a single step in Python where it detects the host OS.

Feel free to close though if you don't want to do that.

makubacki (approved):

Steps/UploadCodeCoverage.yml
```diff
@@ -0,0 +1,77 @@
...
+        filename = 'codecov'
+        checksum_url = 'https://uploader.codecov.io/latest/linux/codecov.SHA256SUM'
+        checksum_filename = 'codecov.SHA256SUM'
+        print(f'##vso[task.setvariable variable=codecov_uploader_cmd]./{filename}')
+      else:
+        print(f'##[error]Unsupported Host System! System = {system}.')
+        print(f'##vso[task.complete result=Failed;]Unsupported Host System! System = {system}.')
+
+      response = requests.get(url)
+      if response.status_code == 200:
```
makubacki, line 51: If the response getting the application is not successful, is there a point in continuing to checksum verification?

makubacki: Similar for the other response below.

Steps/UploadCodeCoverage.yml
```diff
@@ -0,0 +1,77 @@
...
+      if expected_hash != actual_hash.hexdigest():
+        print(f'##[error]Checksum did not match. Expected: {expected_hash}; Actual: {actual_hash.hexdigest()}.')
+        print(f'##vso[task.complete result=Failed;]Hash Mismatch.')
+
+      if system == 'Linux':
+        os.chmod(filename, 0o755)
+
+- script: |
+    $(codecov_uploader_cmd) -t ${{ parameters.codecov_token }} -s ${{ parameters.report_dir }} -Z
+  displayName: Upload Code Coverage to Codecov.io
```
makubacki: Newline at eof.

Steps/UploadCodeCoverage.yml
Javagedes: I did move to python.

Steps/UploadCodeCoverage.yml
Javagedes: Added the print statements to change the result to failed as requested when the request response was not 200. Also added the particular status code to the statement.

Steps/UploadCodeCoverage.yml
```diff
@@ -0,0 +1,83 @@
...
+  default: ''
+
+steps:
+- task: PythonScript@0
+  displayName: Download and Verify Codecov Uploader
+  inputs:
+    scriptSource: inline
+    script: |
+      import platform
+      import requests
```
apop5, line 28: what is the pip module that installs requests? do we need to add it to pip-requirements?

apop5 (approved):

Steps/UploadCodeCoverage.yml
Javagedes: I've updated it such that it will automatically install the latest version of requests. It can also be turned off by using the parameter `install_dependencies: false`

PR #266, opened 2023-11-02
Update Code Coverage Calculations

Javagedes:
Update code coverage calculations to have the ability to reorganize the report by INF using `stuart_parse` and `stuart_report.`

I have confirmed this as working using MuDevOpsWrapper for MU_BASECORE PRs:

Windows: [Pipelines - Run 20231102.11 (azure.com)](https://dev.azure.com/projectmu/mu/_build/results?buildId=57362&view=codecoverage-tab)
Linux: [Pipelines - Run 20231102.11 (azure.com)](https://dev.azure.com/projectmu/mu/_build/results?buildId=57363&view=codecoverage-tab)

PR #270, opened 2023-11-09
MuDevOpsWrapper: Allow coverage uploads to ado or codecov.io

Javagedes:
## Description

This change reverts the changes made in #266 in favor of performing the reformatting in the HostBasedUnitTestRunner plugin. From there it allows the consumer of `MuDevOpsWrapper` to configure where code coverage data is uploaded to, either `ado` (The azure devops pipeline that is running) or `codecov` (The codecov account associated with the repository).

* Uploading to ado remains the same - A final job runs that merges coverage data and uploads it to azure pipelines
* Due to the advanced capabilities, uploading to codecov runs per matrix job and uploads using flags to separate code coverage data between packages.

## Integration Instructions

The previous logic remains the default, that is to say code coverage is uploaded to the azure devops pipeline that is running.  Should a consumer set `coverage_publish_target` to 'codecov', they must ensure that `CODECOV_TOKEN` is set as a private variable in the pipeline.


makubacki:
Can't get to this right now, but I'll come back soon.

makubacki (approved):

.sync/Version.njk
```diff
@@ -30,7 +30,7 @@
 #}
 
 {# The git ref value that files dependent on this repo will use. #}
-{% set mu_devops = "v7.1.0" %}
+{% set mu_devops = "v7.2.0" %}
```
apop5, line 33: is this going to break until 7.2.0 is released?

apop5 (approved):

.sync/Version.njk
makubacki: No. The version will roll to 7.2.0 when this is merged due to the `type:feature-request` label being present.

apop5: I see the plan, merge and immediately release then let it be mirrored over to the repos.

makubacki: If another PR unexpectedly rolls the same version component before this one it would throw off the version predicted here, but there's not a risk of that happening at the moment.

PR #273, opened 2023-11-15
Jobs/PrGate.yml: Update Code Coverage commands

Javagedes:
Updates the Code coverage command to also use the following flags:

CC_FLATTEN=TRUE - de-duplicates source file coverage due to the same source file being used by multiple INFs.

CC_FULL=TRUE - inserts coverage data (correct code line count, but zero lines covered) for all source files in the package that are not present in the original coverage report.

## Integration Instructions

Pipelines consuming this change must add pygount to their pip-requirements file.

### DIFF
--- a/Jobs/PrGate.yml
+++ b/Jobs/PrGate.yml
@@ -177,7 +177,7 @@ jobs:
         ${{ else }}:
           self_host_agent: false
         ${{ if eq(parameters.calculate_code_coverage, true) }}:
-          extra_build_args: CODE_COVERAGE=TRUE
+          extra_build_args: CODE_COVERAGE=TRUE CC_FLATTEN=TRUE CC_FULL=TRUE
         ${{ else }}:
           extra_build_args: ''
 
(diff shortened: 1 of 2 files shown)


Jobs/PrGate.yml
makubacki, line 1: You'll need a version roll for this, right?

Javagedes: It does, but I'm not in as big of a rush to get this in, so I did not plan on immediately rolling the version - unless there is nothing else planned. If that the case, I can roll the version in this PR too

makubacki: In that case it sounds like you could hold the PR and then roll the version before it's completed. It doesn't really matter but I wouldn't expect much else to come in that I'm aware of before this would be picked up.

Javagedes: I have added the version.njk roll to 8.0.0!

Thanks

makubacki (approved):

apop5 (approved):

cfernald (approved):

PR #274, opened 2023-11-15
Steps/UploadCodeCoverage.yml: Remove dependency and conditionalize [Rebase & FF]

makubacki:
Contains two changes related to uploading code coverage:

1. Remove edk2toollib dependency

   Some repos (like pure Rust repos) are expected not to depend on
   pytools. Since the codecov application can simply be run without
   `RunCmd()`, do that and prevent the need for those repos to depend
   on edk2toollib or unnecessary logic in the pipelines to bring it in.

2. Conditionalize coverage upload steps on codecov token presence

   Some repos may not upload to codecov for various reasons. Those repos
   won't set the code coverage token so conditionalize the upload steps
   on the token.

---

Will unblock https://github.com/microsoft/mu_rust_hid/pull/12

### DIFF
--- a/Steps/UploadCodeCoverage.yml
+++ b/Steps/UploadCodeCoverage.yml
@@ -101,17 +101,25 @@ steps:
     scriptSource: inline
     script: |
       from pathlib import Path
-      from edk2toollib.utility_functions import RunCmd
       import io
       import os
+      import subprocess
 
       COV_FLAG = os.environ['COV_FLAG']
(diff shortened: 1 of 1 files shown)


Javagedes (approved):

kuqin12 (approved):

makubacki:
Merging. Tested in https://github.com/microsoft/mu_basecore/pull/625.

PR #277, opened 2023-11-17
Multiple Updates to code coverage

### LINKED ISSUE #267: Optimize code coverage publishing in pipelines
When a large number of pipeline runs at once (e.g. pytools pip update), the `PublishCoverage` job in [`Jobs/PrGate.yml`](https://github.com/microsoft/mu_devops/blob/main/Jobs/PrGate.yml#L169) significantly contributes to the overall time for the pipeline to finish.

This makes PRs take much longer to reach a status check finished state and for the overall job queue to drain. Essentially, the pipeline has to queue twice, once for all of the matrix jobs and again for the code coverage publishing job.

This issue tracks optimizing code coverage publishing to reduce the impact on overall pipeline execution time.

For example, a matrix will spawn `N` jobs. At the end of each matrix job checking if all other matrix jobs (`N-1`) are complete, perhaps using the ADO REST API, and then attempting to publish code coverage directly from the job, etc.

### LINKED ISSUE #276: Fix codecov upload command parameters
In 61bd5b59226d7d2eafae17fcab72d43c0bfc5806, I originally passed the command and params in a sequence to `Popen()`. I switched `params` to be a single string instead of a sequence. I forgot to update the `Popen()` call to pass the value as a single string instead of the sequence.

The command will still work but parameters will not be passed correctly until only the single string is passed.

@Javagedes will include this in another change he's making.

Javagedes:
This PR contains multiple updates to code coverage when executed via MuDevOpsWrapper.yml

1. Removed calculate_code_coverage - This was redundant, as `coverage_publish_target` could be used instead (by having it not set, or selecting codecov / ado
2. Updates all usage of `coverage_publish_target` to be condition of a step, rather than used in a compile time expression for adding / removing steps. This is to support setting the parameter via pipeline variables. This has the unfortunate side effect of the steps always being present in the pipeline build, just skipped, which is not as clean as before.
3. Updates ado code coverage uploads to be per-matrix-job by utilizing `PublishCodeCoverageResults@2` which can merge code coverage results. This resolves #267
4. Resolves #276 by passing the entire command to `subprocess.Popen` and passing `shell=True` to the same.
5. Adds information to the base Readme on code coverage.

[Build](https://dev.azure.com/projectmu/mu/_build/results?buildId=58501&view=results) verifies uploading to codecov works as expected
[Build](https://dev.azure.com/projectmu/mu/_build/results?buildId=58500&view=results) verifies uploading to ado works as expected
[Build](https://dev.azure.com/projectmu/mu/_build/results?buildId=58535&view=results) verifies not specifying a coverage target works as expected

# Breaking Change

`calculate_code_coverage` must be removed from any yaml file extending MuDevOpsWrapper.yml

### DIFF
--- a/.sync/azure_pipelines/MuDevOpsWrapper.yml
+++ b/.sync/azure_pipelines/MuDevOpsWrapper.yml
@@ -1,5 +1,5 @@
 ## @file
-# Azure Pipeline build file for a build using mu_devops..
+# Azure Pipeline build file for a build using mu_devops.
 #
 # NOTE: This file is automatically synchronized from Mu DevOps. Update the original file there
 #       instead of the file in this repo.
--- a/.sync/azure_pipelines/MuDevOpsWrapper.yml
+++ b/.sync/azure_pipelines/MuDevOpsWrapper.yml
@@ -1,5 +1,5 @@
 ## @file
-# Azure Pipeline build file for a build using mu_devops.
+# Azure Pipeline build file for a build using mu_devops..
 #
 # NOTE: This file is automatically synchronized from Mu DevOps. Update the original file there
 #       instead of the file in this repo.
--- a/.sync/azure_pipelines/MuDevOpsWrapper.yml
+++ b/.sync/azure_pipelines/MuDevOpsWrapper.yml
@@ -93,7 +93,6 @@ jobs:
     do_ci_build: ${{ parameters.do_ci_build }}
     do_ci_setup: ${{ parameters.do_ci_setup }}
     do_pr_eval: ${{ parameters.do_pr_eval }}
-    calculate_code_coverage: ${{ parameters.calculate_code_coverage }}
     coverage_publish_target: ${{ parameters.coverage_publish_target }}
     do_non_ci_setup: ${{ parameters.do_non_ci_setup }}
     do_non_ci_build: ${{ parameters.do_non_ci_build }}
--- a/.sync/azure_pipelines/MuDevOpsWrapper.yml
+++ b/.sync/azure_pipelines/MuDevOpsWrapper.yml
@@ -45,7 +45,7 @@ parameters:
 - name: coverage_publish_target
   displayName: Code Coverage Publish Target
   type: string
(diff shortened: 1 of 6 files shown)


.sync/azure_pipelines/MuDevOpsWrapper.yml
```diff
@@ -42,13 +42,10 @@ parameters:
   displayName: Perform Stuart PR Evaluation
   type: boolean
   default: true
-- name: calculate_code_coverage
-  displayName: Calculate Code Coverage From Unit Tests
-  default: false
 - name: coverage_publish_target
   displayName: Code Coverage Publish Target
   type: string
```
makubacki, line 47: It seems a little more clear to add the allowed values under the `values` keyword.

```suggestion
- name: coverage_publish_target
  displayName: Code Coverage Publish Target
  type: string
  values:
  - 'ado'
  - 'codecov'
  - ''
```

makubacki (approved):

Steps/UploadCodeCoverage.yml
```diff
@@ -109,13 +143,14 @@ steps:
...
         if COV_FLAG:
-          params += f' -F {COV_FLAG}'
+          cmd += f' -F {COV_FLAG}'
         process = subprocess.Popen(
-          [UPLOAD_CMD, params],
+          cmd,
           stdout=subprocess.PIPE,
-          stderr=subprocess.PIPE)
+          stderr=subprocess.PIPE,
+          shell=True)
```
makubacki, line 153: Generally, you should avoid using the shell. Was that needed? I thought everything would be resolved going into the final string.

makubacki: Nvm, I think you're depending on that to implicitly get the token.

makubacki: The problem (openness) and simplicity of env vars.

kuqin12 (approved):

apop5 (approved):

PR #279, opened 2023-11-20
MuDevOpsWrapper.yml: Bugfix bad value options

Javagedes:
the `values` config does not work as variables are not evaluated until the step runs, so what actually gets passed and verified against the `values` config ends up being the string `$(variable_name)`, which always fails as we were saying the value could only be "ado" or "codecov" or ""

This PR updates all yaml files to no longer pass the environment variable as a parameter from `MuDevOpsWrapper` -> `Jobs/PrGate` -> `Steps/PrGate` -> `UploadCodeCoverage` and instead uses the environment variable directly in `UploadCodeCoverage` All aforementioned yaml files have been updated with documentation on how to generate code coverage.

### DIFF
--- a/.sync/azure_pipelines/MuDevOpsWrapper.yml
+++ b/.sync/azure_pipelines/MuDevOpsWrapper.yml
@@ -45,7 +45,7 @@ parameters:
 - name: coverage_publish_target
   displayName: Code Coverage Publish Target
   type: string
-  default: 'ado', 'codecov', ''
+  default: 'ado' # 'ado', 'codecov', ''
 - name: container_build
   displayName: Flag for whether this repo should do stuart_setup
   type: boolean
--- a/.sync/azure_pipelines/MuDevOpsWrapper.yml
+++ b/.sync/azure_pipelines/MuDevOpsWrapper.yml
@@ -49,7 +49,7 @@ parameters:
   - 'ado'
   - 'codecov'
   - 'none'
-  default: 'ado' # 'ado', 'codecov', ''
+  default: 'ado'
 - name: container_build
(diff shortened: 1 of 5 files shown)


apop5 (approved):

.sync/azure_pipelines/MuDevOpsWrapper.yml
makubacki, line 1: To make sure I understand correctly, `coverage_publish_target` is currently being passed as:

- From: `MuDevOpsWrapper.yml`: `coverage_publish_target`
- To: `Jobs/PrGate.yml`: `coverage_publish_target`
- To: `Steps/PrGate.yml`: `coverage_publish_target`
- To: `Steps/UploadCodeCoverage.yml`: `upload_target`

Where in `Steps/UploadCodeCoverage.yml`, the `upload_target` parameter is used, for example:

```yaml
- task: PythonScript@0
  displayName: Detect Code Coverage Target and Files
  env:
    UPLOAD_TARGET: ${{ parameters.upload_target }}
    REPORT_DIR: ${{ parameters.report_dir }}
```

And the `upload_target` variable value is used, for example:

```yaml
  condition:  and(eq(variables['upload_target'] , 'ado'), gt(variables.coverage_file_count, 0))
```

So, the value is not expanded at the time of evaluating against the allowed `values` in `MuDevOpsWrapper.yml`, but it is correct (the variable is ready to expand) at the time it is set to the environment variable?

makubacki: We spoke offline. Joey will just use the variable directly in `UploadCodeCoverage.yml` and drop the parameters.

kuqin12 (approved):

PR #288, opened 2023-12-11
Bump setup-python action from v4 to v5 in sync files

makubacki:
Updates the sync files to use v5 so they will match the latest files
updated by dependabot in the synced repo. Also updates the version in
the file in the submodule-release-updater action.


makubacki:
Pushing through since file sync is now syncing old versions without this change.

PR #296, opened 2024-01-18
Change NuGetAuthenticate task to version 1.

### LINKED ISSUE #289: NuGetAuthenticate@0 is deprecated. Switch to use NuGetAuthenticate@1
Need to switch mu_devops to use NuGetAuthenticate@1 tasks.

https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/nuget-authenticate-v0?view=azure-pipelines

https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/nuget-authenticate-v1?view=azure-pipelines

apop5:
Need to switch mu_devops to use NuGetAuthenticate@1 tasks. NuGetAuthenticate@0 is being depreciated, in some pipelines, starting 2024.01.31.

https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/nuget-authenticate-v0?view=azure-pipelines
https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/nuget-authenticate-v1?view=azure-pipelines

Fixes #289

### DIFF
--- a/Steps/NuGet.yml
+++ b/Steps/NuGet.yml
@@ -9,47 +9,47 @@
 ##
 
 parameters:
-  - name: artifacts_identifier
-    displayName: Artifacts Identifier
-    type: string
-    default: "Artifacts"
-  - name: custom_nuget_prep_step_list
-    displayName: Custom NuGet Package Step List
-    type: stepList
-    default:
-      - script: echo No custom NuGet prep steps provided
-  - name: include_binaries
-    displayName: Include build binaries in the Nuget package
-    type: boolean
-    default: true
-  - name: include_logs
-    displayName: Include build logs in the Nuget package
-    type: boolean
-    default: false
-  - name: include_other
-    displayName: Include other build artifacts in the Nuget package
-    type: boolean
-    default: false
-  - name: license_file_path
-    displayName: License File Path (Absolute Path)
-    type: string
-    default: ""
+- name: artifacts_identifier
+  displayName: Artifacts Identifier
+  type: string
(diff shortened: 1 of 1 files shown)


apop5:
@makubacki 

Do you know of a pipeline where Steps/Nuget.yml is being used so I can test this?

makubacki:
Was there a particular reason to change the YAML formatting? If that's going to be done, it should be in a separate commit from the actual change. However, if that is a direction we'd strongly like to go, it really should be done in a single commit in a separate PR (that can be squashed if needed). Then, the formatting only commit can be removed from git blame.

I don't recall NuGet usage without looking around a bit. In some places, like mu_tiano_platforms, that push to a NuGet feed but I don't recall where the template is used versus inline NuGet code.

apop5:
> Was there a particular reason to change the YAML formatting? If that's going to be done, it should be in a separate commit from the actual change. However, if that is a direction we'd strongly like to go, it really should be done in a single commit in a separate PR (that can be squashed if needed). Then, the formatting only commit can be removed from git blame.
> 
> I don't recall NuGet usage without looking around a bit. In some places, like mu_tiano_platforms, that push to a NuGet feed but I don't recall where the template is used versus inline NuGet code.

The formatting was unintentional, I will revert it.

The other repos using nugetauthenticate, when they were updated, did not need any other change from the @0 to @1, but they were doing their authenticates inline in the projects.

makubacki (approved):

Javagedes (approved):

PR #302, opened 2024-01-23
.sync/workflows/codeql: Always remove plugins in .pytools

kenlautner:
With the CodeQL plugin moving to BaseTools (from .pytool) starting in release/202311, update the workflow to always remove unnecessary plugins (that slow down the workflow) in .pytools as opposed to relative the CodeQL plugin path.

### DIFF
--- a/.sync/workflows/leaf/codeql.yml
+++ b/.sync/workflows/leaf/codeql.yml
@@ -321,19 +321,41 @@ jobs:
       if: steps.codeqlcli_cache.outputs.cache-hit != 'true'
       run: stuart_update -c .pytool/CISettings.py -t DEBUG -a ${{ matrix.archs }} TOOL_CHAIN_TAG=${{ matrix.tool_chain_tag }} --codeql
 
+    - name: Find pytool Plugin Directory
+      id: find_pytool_dir
+      shell: python
+      run: |
+        import os
+        import sys
+        from pathlib import Path
+
+        # Find the plugin directory that contains the Compiler plugin
+        plugin_dir = list(Path(os.environ['GITHUB_WORKSPACE']).rglob('.pytool/Plugin/CompilerPlugin'))
+
+        # This should only be found once
+        if len(plugin_dir) == 1:
+            # If the directory is found get the parent Plugin directory
+            plugin_dir = str(plugin_dir[0].parent)
+
+            with open(os.environ['GITHUB_OUTPUT'], 'a') as fh:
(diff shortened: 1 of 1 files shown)


makubacki:
@kenlautner, I updated the title/description to reduce confusion in the future. No one is broken except the 2311 branch which is the cause of the break. So, although this was found late in testing, it is an expected/routine change for moving the plugin location in Project Mu.

.sync/workflows/leaf/codeql.yml
```diff
@@ -321,19 +321,41 @@ jobs:
...
+        # Find the plugin directory that contains the Compiler plugin
+        plugin_dir = list(Path(os.environ['GITHUB_WORKSPACE']).rglob('.pytool/Plugin/CompilerPlugin'))
+
+        # This should only be found once
+        if len(plugin_dir) == 1:
+            # If the directory is found get the parent Plugin directory
+            plugin_dir = str(plugin_dir[0].parent)
+
+            with open(os.environ['GITHUB_OUTPUT'], 'a') as fh:
+                print(f'codeql_plugin_dir={plugin_dir}', file=fh)
```
makubacki, line 341: This will overwrite the valid CodeQL plugin location previously stored.

```suggestion
```

.sync/workflows/leaf/codeql.yml
```diff
@@ -321,19 +321,41 @@ jobs:
...
+        else:
+            print("::error title=Workspace Error!::Failed to find Mu Basecore plugin directory!")
```
makubacki, line 343: Can you please update this to:

```suggestion
            print("::error title=Workspace Error!::Failed to find Mu Basecore .pytool/Plugin directory!")

```

.sync/workflows/leaf/codeql.yml
makubacki: Looks like you're using the output later as `pytool_plugin_dir`. So:

```suggestion
            with open(os.environ['GITHUB_OUTPUT'], 'a') as fh:
                print(f'pytool_plugin_dir={plugin_dir}', file=fh)
```

kenlautner: You're correct.  I meant to change this to be `pytool_plugin_dir` which I'll fix now.

.sync/workflows/leaf/codeql.yml
kenlautner: Updated.

makubacki (approved):

Javagedes (approved):

PR #304, opened 2024-01-24
pip: Sync core stuart requirements

Javagedes:
**Note**: I plan on this being new functionality for 202311 and should not be merged (if accepted) until then.

## Description

There are multiple core pip packages required for stuart to work (both for the invocable itself, and the core plugins).  This commit adds a requirements.txt file to the .pytool directory to sync these requirements across all repositories using stuart.

Repositories should not pip install this file directly, but instead link against it in their own requirements file using `-r .pytool/requirements.txt`. They should still add repository specific requirements to their own requirements file.

Adds packages to the ignore list of dependabot so that it does not create PRs for dependencies in `.pytool/requirements.txt`

The main benefit of this change is to allow multiple requirements to be updated at the same time, in a single PR, to reduce churn and total PRs necessary.

## Example

mu_feature_config
```
.
├── .pytool
│      ├── requirements.txt (sync'd)
├── pip-requirements.txt
```

pip-requirements.txt
```
-r .pytool/requirements.txt
pywin32==306; sys_platform == 'win32
```

final command (unchanged)
`pip install --upgrade -r pip-requirements.txt`

### DIFF
--- a/.sync/Files.yml
+++ b/.sync/Files.yml
@@ -621,3 +621,21 @@ group:
       microsoft/mu_basecore
       microsoft/mu_plus
       microsoft/mu_tiano_platforms
+
+# Pip - Stuart pip requirements
+  - files:
+    - source: .sync/pytool/requirements.txt
+      dest: .pytool/requirements.txt
+    repos: |
+      microsoft/mu_basecore
+      microsoft/mu_common_intel_min_platform
+      microsoft/mu_feature_dfci
+      microsoft/mu_plus
+      microsoft/mu_silicon_arm_tiano
+      microsoft/mu_silicon_intel_tiano
+      microsoft/mu_tiano_platforms
+      microsoft/mu_tiano_plus
+      microsoft/mu_feature_config
+      microsoft/mu_oem_sample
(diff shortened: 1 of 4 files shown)


.sync/Files.yml
makubacki, line 1: Leaving this comment open to prevent the PR from being merged before 2311 syncing is ready.

I'll come back to the PR then.

.sync/pytool/requirements.txt
```diff
@@ -0,0 +1,28 @@
...
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+# https://pypi.org/project/pip/
+# https://pip.pypa.io/en/stable/user_guide/#requirements-files
+# https://pip.pypa.io/en/stable/reference/pip_install/#requirements-file-format
+# https://www.python.org/dev/peps/pep-0440/#version-specifiers
+##
+edk2-pytool-library==0.20.0
+edk2-pytool-extensions~=0.27.0
+edk2-basetools==0.1.49
```
Javagedes, line 23:
```suggestion
```

PR #315, opened 2024-02-06
Update container version in Version.njk

Javagedes:
Updates the container version to the newly built container, which uses rust toolchain version 1.74.0.


makubacki:
Closing since it is a duplicate with https://github.com/microsoft/mu_devops/pull/316.

PR #318, opened 2024-02-15
Add pipeline to cache cargo binstall

Javagedes:
Adds a pipeline to cache cargo binstall.

Adds `.tgz` (short for `.tar.gz`) support to the DownloadCargoBinaryFromGitHub.py script. Also fixes a logic error in the conditional that decides which assets to download.

See successful pipeline run: https://dev.azure.com/projectmu/mu/_build/results?buildId=64211&view=logs&j=402df595-8a03-57d0-41e1-0654a1932349

### DIFF
--- a/Scripts/DownloadCargoBinaryFromGitHub/DownloadCargoBinaryFromGitHub.py
+++ b/Scripts/DownloadCargoBinaryFromGitHub/DownloadCargoBinaryFromGitHub.py
@@ -44,7 +44,7 @@ if len(releases) == 0:
 # Download assets
 for asset in releases[0]['assets']:
     name = asset['name'].lower()
-    if (("x86_64-pc-windows-msvc.full" in name or "x86_64-unknown-linux-gnu.full" in name)
+    if (("x86_64-pc-windows-msvc" in name or "x86_64-unknown-linux-gnu" in name)
        and asset['name'].endswith(('.zip', '.tar.gz', '.tgz'))):
         filepath = DOWNLOAD_DIR / asset['name']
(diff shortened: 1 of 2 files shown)


Scripts/DownloadCargoBinaryFromGitHub/DownloadCargoBinaryFromGitHub.py
```diff
@@ -44,8 +44,8 @@
 # Download assets
 for asset in releases[0]['assets']:
     name = asset['name'].lower()
-    if ("x86_64-pc-windows-msvc" in name or "x86_64-unknown-linux-gnu" in name
-       and asset['name'].endswith(('.zip', '.tar.gz'))):
+    if (("x86_64-pc-windows-msvc.full" in name or "x86_64-unknown-linux-gnu.full" in name)
```
makubacki, line 47: Is `".full"` expected to work with all releases?

makubacki (approved):

Scripts/DownloadCargoBinaryFromGitHub/DownloadCargoBinaryFromGitHub.py
Javagedes: Updated - I pushed the commit that removed this to the wrong remote. locally, it works fine without the ".full". Running it in pipeline format with ".full" removed, just to be sure.

Javagedes: Pipeline was successful.

apop5 (approved):

PR #319, opened 2024-02-16
Use cargo-binstall to sync rust tool versions

Javagedes:
Update RustSetupSteps to download cargo-binstall and use it to download all other rust tools. This is to make it easy for CI to pass the updated RustEnvironmentCheck (https://github.com/microsoft/mu_basecore/pull/737) by downloading the exact tool versions as specified in the rust-toolchain.toml file.

### DIFF
--- a/.sync/Version.njk
+++ b/.sync/Version.njk
@@ -30,7 +30,7 @@
 #}
 
 {# The git ref value that files dependent on this repo will use. #}
-{% set mu_devops = "v9.1.3" %}
+{% set mu_devops = "v9.1.5" %}
 
 {# The latest Project Mu release branch value. #}
 {% set latest_mu_release_branch = "release/202311" %}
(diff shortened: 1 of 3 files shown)


makubacki (approved):
One of the reasons for caching these in local pipelines was to reduce dependencies on external GitHub resources. But I think stability has been good enough lately that might be less of a concern now.

Javagedes:
> One of the reasons for caching these in local pipelines was to reduce dependencies on external GitHub resources. But I think stability has been good enough lately that might be less of a concern now.

Do you want this moved to something else then?

makubacki:
> > One of the reasons for caching these in local pipelines was to reduce dependencies on external GitHub resources. But I think stability has been good enough lately that might be less of a concern now.
> 
> Do you want this moved to something else then? Or is this referencing the fact that binstall works by downloading from a gh release?

Just the fact that it installs from GitHub releases.

apop5 (approved):

PR #332, opened 2024-04-11
Steps/SetupGitConfig.yml: Set auto.crlf to false in builds

makubacki:
Pipelines that build firmware come in through two entry points:

1. Steps/BuildPlatform.yml - Platform build (e.g. mu_tiano_platforms)
2. Steps/PrGate.yml - Core build (e.g. mu_basecore)

This change updates both paths to set core.autocrlf to false so line
endings are checked out as-is on Windows. This allows checks like the
LineEndingCheck to have similar results to Linux.

The change is placed in a dedicated file to make it easier to group
similar git config tweaks that might be required in pipelines for
flows that build firmware code in the future.

### DIFF
--- a/Steps/BuildPlatform.yml
+++ b/Steps/BuildPlatform.yml
@@ -75,6 +75,8 @@ steps:
     # Note: Depth cannot be limited if PR Eval is used. A pipeline may choose
     #       to use a shallow checkout if PR eval is not used.
 
+- template: SetupGitConfig.yml
+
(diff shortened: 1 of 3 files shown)


makubacki:
Waiting for results from these tests before publishing:

- [Should fail](https://dev.azure.com/projectmu/mu/_build/results?buildId=67226&view=results)
- [Should pass](https://dev.azure.com/projectmu/mu/_build/results?buildId=67227&view=results)

PR #333, opened 2024-04-12
Prevent line ending conversion in repos

makubacki:
Unsets the text attribute on all files to prevent their line endings
from being normalized.

See https://git-scm.com/docs/gitattributes for background.

This was previously accomplished solely by instructing users to set
core.autocrlf to false. However, that does not translate to pipelines
and setting in a pipeline requires the system level git configuration
to be modified prior to checkout. Git configs cannot be checked into
to a repo and automatically used for security reasons.

This is a simple way to check the change into the repo so it applies
for all users including pipelines to prevent line ending conversion.

This allows plugins like LineEndingCheck to produce similar results
in pipelines on Linux and Windows agents.


makubacki:
Tested in the following mu_basecore pipeline runs:

- [Expected Failure](https://dev.azure.com/projectmu/mu/_build/results?buildId=67237&view=results)
- [Expected Pass](https://dev.azure.com/projectmu/mu/_build/results?buildId=67238&view=results)

apop5 (approved):

os-d (approved):

makubacki:
On hold to let this PR verify there's no unexpected impact before rolling out to other repos.

https://github.com/microsoft/mu_basecore/pull/809

makubacki:
> On hold to let this PR verify there's no unexpected impact before rolling out to other repos.
> 
> [microsoft/mu_basecore#809](https://github.com/microsoft/mu_basecore/pull/809)

No negative consequences seen from that PR, so I'm going to merge this PR now.

PR #337, opened 2024-04-12
Version.njk: Update Ubuntu 22 container for Rust 1.76.0

makubacki:
Only change is building with Rust 1.76.0 instead of Rust 1.74.0.


makubacki:
I forgot the container image needs to be updated in the last commit that rolled out new Mu DevOps version. Any pipelines that require it should have failed and will be amended after this PR completes.

Javagedes (approved):

kenlautner (approved):

kuqin12 (approved):

PR #340, opened 2024-05-01
UploadCodeCoverage.yml: Add auto retry 

apop5:
Infrequently, pipelines have been failing when trying to communicate to codecov.io

Adding 5 retry attempts to the initial communication.

### DIFF
--- a/Steps/UploadCodeCoverage.yml
+++ b/Steps/UploadCodeCoverage.yml
@@ -164,7 +164,7 @@ steps:
           output, error = process.communicate()
           print(f"##[debug]{output.decode('utf-8')}")
           if process.returncode == 0:
-            continue
+            break
           time.sleep(.5*loop)
 
         if loop == 4 and process.returncode != 0:
--- a/Steps/UploadCodeCoverage.yml
+++ b/Steps/UploadCodeCoverage.yml
@@ -94,7 +94,7 @@ steps:
       
       session = requests.Session()
       retries = Retry(total=5,
-                backoff_factor=1,
+                backoff_factor=.5,
                 status_forcelist=[ 500, 502, 503, 504 ])
       session.mount('https://', HTTPAdapter(max_retries=retries))
 
@@ -143,6 +143,7 @@ steps:
       import io
       import os
       import subprocess
+      import time
 
       COV_FLAG = os.environ['COV_FLAG']
       REPORT_DIR = os.environ['REPORT_DIR']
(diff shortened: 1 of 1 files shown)


Steps/UploadCodeCoverage.yml
```diff
@@ -145,14 +152,20 @@ steps:
...
-          cmd,
-          stdout=subprocess.PIPE,
-          stderr=subprocess.PIPE,
-          shell=True)
-        output, error = process.communicate()
-        print(f"##[debug]{output.decode('utf-8')}")
+        loop = 0
+        process = 0
+        while loop < 5 and process == 0:
+          process = subprocess.Popen(
```
Javagedes, line 158: we probably want to add a delay, similar to the `Retry()` code you added above. The `Retry()` code doubles the delay time between retries each iteration due to the backoff_factor of 1.

Steps/UploadCodeCoverage.yml
```diff
@@ -145,14 +152,20 @@ steps:
...
-        process = subprocess.Popen(
-          cmd,
-          stdout=subprocess.PIPE,
-          stderr=subprocess.PIPE,
-          shell=True)
-        output, error = process.communicate()
-        print(f"##[debug]{output.decode('utf-8')}")
+        loop = 0
+        process = 0
+        while loop < 5 and process == 0:
```
makubacki, line 157: I think this would be a little cleaner with a `for ... in range()` loop where `5` is replaced with a named constant.

makubacki (approved):

Steps/UploadCodeCoverage.yml
apop5: Added code and used time.sleep. 

Matched documented method of [retry](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.retry)

Javagedes (approved):

Steps/UploadCodeCoverage.yml
```diff
@@ -145,14 +153,22 @@ steps:
...
+        for loop in range(0,5):
+          process = subprocess.Popen(
+            cmd,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            shell=True)
+          output, error = process.communicate()
+          print(f"##[debug]{output.decode('utf-8')}")
+          if process.returncode == 0:
+            return
```
Javagedes, line 167: This needs to be a break

Javagedes: L152 we are looping through and submitting multiple reports. This return will cause an exit after the first report is submitted.

Javagedes (changes requested):
The `return` causes the script to stop after submitting the first code coverage report, but this code uploads one per package in the same execution.

Javagedes (approved):

PR #342, opened 2024-05-15
Steps/BuildPlatform.yml & Steps/PrGate.yml: Add parameter to install pip modules

pharlikar:
SetupPythonPreReqs.yml already has a parameter with default set to true, but the user cannot change it. So, extending the install_pip_modules paramter, so that the user can override it

### DIFF
--- a/Steps/PrGate.yml
+++ b/Steps/PrGate.yml
@@ -72,6 +72,10 @@ parameters:
   displayName: Install Build Tools
   type: boolean
   default: true
+- name: install_pip_modules
+  displayName: Install PIP Modules
+  type: boolean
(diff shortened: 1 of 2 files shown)


Steps/PrGate.yml
makubacki, line 1: Can you please add a similar parameter to `Steps/BuildPlatform.yml` and similarly pass it to `SetupPythonPreReqs.yml` there?

makubacki (approved):

apop5 (approved):

Steps/PrGate.yml
pharlikar: updated

PR #343, opened 2024-05-24
remove `-a` flag on stuart_build commands

Javagedes:
`-a` is not a first class parameter for stuart_build, so mu_devops should not expect that it exists. If a platform has added a necessary `-a` argument, then they can pass it to the command in mu_devops through the `build_flags` for BuildPlatform.yml or ` extra_build_args` for PrGate.yml

### DIFF
--- a/Steps/BuildPlatform.yml
+++ b/Steps/BuildPlatform.yml
@@ -18,10 +18,6 @@ parameters:
   displayName: Other Artifacts to Publish
   type: string
   default: ''
-- name: build_arch
-  displayName: Architectures (e.g. IA32, X64)
-  type: string
-  default: ''
 - name: build_base_tools
   displayName: Build BaseTools
   type: boolean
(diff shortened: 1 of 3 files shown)


os-d (approved):

apop5 (approved):

Steps/BuildPlatform.yml
```diff
@@ -128,14 +128,14 @@ steps:
 - task: CmdLine@2
   displayName: Build
   inputs:
-    script: stuart_build -c ${{ parameters.build_file }} TOOL_CHAIN_TAG=${{ parameters.tool_chain_tag}} TARGET=${{ parameters.build_target}} -a ${{ parameters.build_arch}} ${{ parameters.build_flags}}
+    script: stuart_build -c ${{ parameters.build_file }} TOOL_CHAIN_TAG=${{ parameters.tool_chain_tag}} TARGET=${{ parameters.build_target}} ${{ parameters.build_flags}}
```
makubacki, line 131: I understand the platform can now pass `-a <ARCH>`, but doesn't this leave `parameters.build_arch` unused in this command? That seems unexpected given its a parameter to the template.

Javagedes: @makubacki the setup and update commands still have the -a flag passed to them. I'm not sure how necessary that is. If you don't think that is necessary for the BuildPlatform.yml, I'm happy to remove them there too

makubacki: I'd prefer dropping the parameter entirely and marking this as a breaking change. I think it cleans up how parameters behave.

makubacki (approved):

Steps/BuildPlatform.yml
Javagedes: Dropped the parameter and specified this as a breaking change.

Javagedes: I will roll this as a 10.0.0 release.

PR #346, opened 2024-06-05
Update commit update logic

kuqin12:
The existing logic will not update the tag if a non-tag commit is being picked up.

The change will fix the issue by comparing the current commit and the tag commit to ensure that the primary repo is actually sitting on the tag commit, otherwise a submodule update PR will be created.


kuqin12:
An example can be seen here: https://github.com/microsoft/mu_tiano_platforms/pull/977

PR #348, opened 2024-06-05
Add CLANGPDB prerequisite steps template

kuqin12:
This change added the prerequisite template for installing CLANGPDB at designated locations.

This is needed so that "make" does not complain about the clang install path has space in it.


Steps/SetupToolChainTagPreReqs.yml
```diff
@@ -0,0 +1,25 @@
+## @file
+# Azure Pipelines step to setup Python pre-requisites.
+#
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+parameters:
```
kuqin12, line 8: I guess I do not need this..

cfernald (approved):

Steps/SetupToolChainTagPreReqs.yml
kuqin12: Removed

apop5 (approved):

PR #349, opened 2024-06-26
.sync/codeql: Add retry on GitHub REST Access Failures

makubacki:
Attempts sometimes fail on the first try and later succeed when
reaching out to the GitHub REST API. To prevent actions from having
to be rerun, try up to 5 times waiting 10 seconds between each try.

### DIFF
--- a/.sync/workflows/leaf/codeql-platform.yml
+++ b/.sync/workflows/leaf/codeql-platform.yml
@@ -159,11 +159,21 @@ jobs:
         import os
         import requests
         import sys
+        import time
+
+        def get_response_with_retries(url, retries=5, wait_time=10):
+          for attempt in range(retries):
+            response = requests.get(url)
+            if response.status_code == 200:
+              return response
+            print(f"::warning title=GitHub API Access Error!::Attempt {attempt + 1} failed. Retrying in {wait_time} seconds...")
+            time.sleep(wait_time)
+          return response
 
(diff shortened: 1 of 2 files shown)


makubacki:
Tested in https://github.com/makubacki/mu_basecore/actions/runs/9685929840

Javagedes (approved):

.sync/workflows/leaf/codeql-platform.yml
```diff
@@ -159,11 +159,21 @@ jobs:
         import os
         import requests
         import sys
+        import time
+
+        def get_response_with_retries(url, retries=5, wait_time=10):
+          for attempt in range(retries):
+            response = requests.get(url)
+            if response.status_code == 200:
```
Javagedes, line 167: technically any code in the 200 range is a success. Do we need to account for any of those?

makubacki: In this case for a GET request the response code is defined as `200`.

- [General REST API HTTP Status Codes](https://gist.github.com/subfuzion/669dfae1d1a27de83e69)
- [Get the combined status for a specific reference](https://docs.github.com/en/rest/commits/statuses?apiVersion=2022-11-28#get-the-combined-status-for-a-specific-reference)

While the others are valid, I wouldn't expect them for this request.

VivianNK (approved):

PR #355, opened 2024-07-22
Steps/BuildPlatform.yml & Steps/PrGate.yml: Introduce publish_artifacts parameter

MarcChen46:
A new publish_artifacts parameter is introduced to the Steps/PrGate.yml and Steps/BuildPlatform.yml templates to allow consuming pipelines to separate publication of artifacts from their copying to the staging directory. This is useful for a consumer to customize publication of the artifact staging directory after artifacts are copied by these templates. The default for publish_artifacts is true for complete backward compatibility with existing pipelines


MarcChen46:
@apop5, please help reviewing it and let me know if there is any feedback, thanks.

Steps/BuildPlatform.yml
```diff
@@ -135,19 +139,29 @@ steps:
   condition: and(and(gt(variables.pkg_count, 0), succeeded()), eq(variables['Run'], true))
   timeoutInMinutes: ${{ parameters.run_timeout }}
 
-# Copy build logs to the artifact staging directory
-- template: CommonLogCopyAndPublish.yml
-  parameters:
-    artifacts_identifier: ${{ parameters.artifacts_identifier }}
+- ${{ if eq(parameters.publish_artifacts, true) }}:
```
makubacki, line 142: In general, I'm fine with the changes. But, there are two important changes in behavior that should be restored.

1. Before, each set of artifacts was published before the copy/publish for the next. The copy and publish tasks are sequenced so that if something went wrong highest priority artifacts (i.e. logs) to figure out what happened would be available. Given this is calling a template where the implementation is unknown, having publication right away was a step to help diagnose problems.
2. This is publishing the contents of the entire `$(Build.ArtifactStagingDirectory)` directory whereas previously individual paths were being published. In consumer pipelines, this might change what is published if contents were sitting in the artifacts directory and not explicitly published. That is an unnecessary change in behavior to accomplish this change.

To solve both of these, please put a publish step after each copy template matching the old publish step where the artifact specific directory is published.

Steps/BinaryCopy.yml
makubacki, line 1: The PR description needs a simple explanation of what was done and how to use it. For example:

> "A new `publish_artifacts` parameter is introduced to the `Steps/PrGate.yml` and `Steps/BuildPlatform.yml` templates to allow consuming pipelines to separate publication of artifacts from their copying to the staging directory. This is useful for a consumer to customize publication of the artifact staging directory after artifacts are copied by these templates. The default for `publish_artifacts` is `true` for complete backward compatibility with existing pipelines."

MarcChen46: updated description per feedback.

Steps/BuildPlatform.yml
MarcChen46: Restore the flow back and only introduce publish_artifacts parameter to control the execution of the artifact publish step.

MarcChen46:
Hi @makubacki, please review it again and let me know if there is any extra feedbacks. thanks a lot.

apop5:
This looks good, Thanks for submitting it.

makubacki (approved):

makubacki:
> Hi @makubacki, please review it again and let me know if there is any extra feedbacks. thanks a lot.

The updates look good. Thanks

PR #356, opened 2024-08-07
Set ImageBase to 0 and Subsystem to EFI Boot Services in config.toml

makubacki:
Currently sets the ImageBase in the PE/COFF header to
`0` as expected for UEFI images.

The cargo config file will be synced to Rust enabled
repos to apply the linker flags regardless of whether
the Rust code is compiled directly or through the
firmware build system.

### DIFF
--- a/.sync/rust_config/config.toml
+++ b/.sync/rust_config/config.toml
@@ -1,0 +1,5 @@
+[target.x86_64-unknown-uefi]
+rustflags = [
+    "-C", "link-arg=/base:0x0",
+    "-C", "link-arg=/subsystem:efi_boot_service_driver",
+]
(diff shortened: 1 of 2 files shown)


.sync/rust_config/config.toml
```diff
@@ -0,0 +1,5 @@
+[target.x86_64-unknown-uefi]
+rustflags = [
+    "-C", "link-arg=/base:0x0",
+    "-C", "link-arg=/subsystem:efi_boot_service_driver",
```
os-d, line 4: nit: you can do a single line with `link-args=/base:0x0 /subsystem:efi_boot_service_driver"`. Don't really care, just if we think that adding future ones is simpler with a single line. That being said, as I type this out I'm changing my mind and I think this is useful to have separated out so that the git blame is easier to identify what came when. So I will just resolve this.

os-d (approved):

vineelko (approved):

vineelko (approved):

.sync/rust_config/config.toml
makubacki: Yeah, agree with keeping them separate.

apop5 (approved):

PR #379, opened 2024-10-15
Add parameter for caller to skip setup or update

kanechen66:
There are cases the pipeline can call BuildPlatform.yml more than 1 time to compile different components in the bios.
In fact, the 2nd or 3rd call, might not require setup/update task.
In order to save pipeline time, this PR creates new parameters to allow the caller to skip setup/update task

### DIFF
--- a/Steps/BuildPlatform.yml
+++ b/Steps/BuildPlatform.yml
@@ -74,6 +74,14 @@ parameters:
   displayName: Publish Artifacts
   type: boolean
   default: true
+- name: skip_setup
+  displayName: Skip Setup
+  type: boolean
+  default: false
+- name: skip_update
+  displayName: Skip Update
+  type: boolean
+  default: false
 
 steps:
 - ${{ if eq(parameters.checkout_self, true) }}:
@@ -109,14 +117,14 @@ steps:
   displayName: Setup
   inputs:
     script: stuart_setup -c ${{ parameters.build_file }} TOOL_CHAIN_TAG=${{ parameters.tool_chain_tag}} -t ${{ parameters.build_target}} ${{ parameters.build_flags}}
-  condition: and(gt(variables.pkg_count, 0), succeeded())
+  condition: and(and(gt(variables.pkg_count, 0), succeeded()), eq('${{ parameters.skip_setup }}', 'false'))
 
 # Stuart Update
 - task: CmdLine@2
   displayName: Update
   inputs:
     script: stuart_update -c ${{ parameters.build_file }} TOOL_CHAIN_TAG=${{ parameters.tool_chain_tag}} -t ${{ parameters.build_target}} ${{ parameters.build_flags}}
-  condition: and(gt(variables.pkg_count, 0), succeeded())
+  condition: and(and(gt(variables.pkg_count, 0), succeeded()), eq('${{ parameters.skip_update }}', 'false'))
 
 # build basetools
 #   do this after setup and update so that code base dependencies


Steps/BuildPlatform.yml
makubacki, line 1: @kanechen66 - How much time is being spent in those steps if they were already run? Typically, they should execute quickly if no work needs to be performed during setup/update.

kanechen66: hi @makubacki 
In one of our proj, the first run is to build fsp and the 2nd run is to build BIOS.
The 2nd run setup/update takes ~1.5+ mins
In 2nd round, from the log it still tries to init/download all the submodules
if i skip these 2 steps, it can save 1.5+ mins in pipeline
thank you

image

makubacki: I think that should be improved separately. This change is okay too.

makubacki (approved):

kanechen66:
hi Michael, 
The UI shows `At least 2 approving reviews are required by reviewers`, however i can't find button allow me to add more reviewers.
Do you know where i can add another reviewer?
thank you :)
image

apop5 (approved):

Steps/BuildPlatform.yml
```diff
@@ -109,14 +117,14 @@ steps:
   displayName: Setup
   inputs:
     script: stuart_setup -c ${{ parameters.build_file }} TOOL_CHAIN_TAG=${{ parameters.tool_chain_tag}} -t ${{ parameters.build_target}} ${{ parameters.build_flags}}
-  condition: and(gt(variables.pkg_count, 0), succeeded())
+  condition: and(and(gt(variables.pkg_count, 0), succeeded()), eq('${{ parameters.skip_setup }}', 'false'))
```
makubacki, line 120: @kanechen66, have you verified the condition against true and false scenarios?

makubacki (approved):
Approving assuming the logic has been exhaustively tested.

apop5:
@makubacki 

would this be a case where using extra_install_step should be used instead of calling platformbuild twice?

i.e.

extra_install_step:
- task: CmdLine@2
          displayName: stuart_build 
          inputs:
            script: stuart_build -c <path-to-platform file>  etc

Steps/BuildPlatform.yml
kanechen66: hi @makubacki 
for `true` case. I tested true case and the setup/update can be skip properly in pipeline
for `false` cae, i tested without setting true from caller.  The setup/update will still be executed in pipeline
image
Is this ok?
thanks

makubacki: Thanks. The logic in pipelines related to booleans can sometimes be unintuitive so I just wanted to check that both cases were tested as this code is written.

makubacki:
> @makubacki
> 
> would this be a case where using extra_install_step should be used instead of calling platformbuild twice?
> 
> i.e.
> 
> extra_install_step:
> 
> * task: CmdLine@2
>   displayName: stuart_build
>   inputs:
>   script: stuart_build -c   etc

That's a good point. @kanechen66 mentioned earlier it is run twice because:

> In one of our proj, the first run is to build fsp and the 2nd run is to build BIOS.

The main steps after the build the template are to "run" the code (if opted into) and then copy the logs, binaries, and other artifacts.

So, @kanechen66, if you did provide the FSP build command as a `extra_install_step` parameter value so that it builds before the platform, would those steps afterwards work for you?

---

Stepping back, we're trying to reduce the total number of parameters, and we already added `extra_install_step` to provide some pre-platform build flexibility.

kanechen66:
sorry I'm a bit confused, did you mean adding extra_install_step in caller can skip the  # Setup repo, # Stuart Update steps?
by the way, in our proj, we already use extra_install_step to call other template.
thank you

makubacki:
> sorry I'm a bit confused, did you mean adding extra_install_step in caller can skip the # Setup repo, # Stuart Update steps? by the way, in our proj, we already use extra_install_step to call other template. thank you

`extra_install_step` is a `stepList` (an arbitrary list of passed in steps) that is executed here in Steps/PlatformBuild.py:

https://github.com/microsoft/mu_devops/blob/93c2bcbcbdf62e9400ba1d2fc69d5e87e9ad9599/Steps/BuildPlatform.yml#L130

The normal build step directly follows it and build the `parameters.build_file` file:

https://github.com/microsoft/mu_devops/blob/93c2bcbcbdf62e9400ba1d2fc69d5e87e9ad9599/Steps/BuildPlatform.yml#L132-L137

In your case can this work?

- `extra_install_step` include the step to build the FSP
  - Note: This will occur after `stuart_setup` and `stuart_update` have completed only one time.
  - Note: This is a `stepList`, so other steps also provided are fine.
- `build_file` be your normal (non-FSP) platform build file (e.g. `PlatformBuild.py`)

kanechen66:
hi @makubacki 
Just tested with the below code. it can skip setup and update. However it's not working because building fsp requires setup/update steps too. 

Here are what I did. let me know if anything wrong.
thank you
1.
remove the original BuildPlatform.yml call for building fsp
2. build fsp in extra steps and 
```
           build_file: xxx/bios_build.py
           ....
           extra_install_step:
                  - template: call_other template
                  - task: CmdLine@2
                    displayName: stuart_build
                    inputs:
                      script: stuart_build -c xxx/platform_build_fsp.py TOOL_CHAIN_TAG=$(tool_chain) TARGET=$(Build.Target) -a $(Build.Arch) ... .....
```

apop5:
> hi @makubacki Just tested with the below code. it can skip setup and update. However it's not working because building fsp requires setup/update steps too.
> 
> Here are what I did. let me know if anything wrong. thank you 1. remove the original BuildPlatform.yml call for building fsp 2. build fsp in extra steps and
> 
> ```
>            build_file: xxx/bios_build.py
>            ....
>            extra_install_step:
>                   - template: call_other template
>                   - task: CmdLine@2
>                     displayName: stuart_build
>                     inputs:
>                       script: stuart_build -c xxx/platform_build_fsp.py TOOL_CHAIN_TAG=$(tool_chain) TARGET=$(Build.Target) -a $(Build.Arch) ... .....
> ```

@kanechen66 

Can you contact me offline to discus this further?

kanechen66:
@apop5 @makubacki 

Sorry, I think i made some mistakes yesterday.
I forgot to remove the skip_setup, skip_update in my test so that the setup/env are never called hence errors.

Just tried again and remove all skip parameters i added. 
Looks like it's working. only one setup/update and fsp then BIOS build
i will test more. thanks for the helps

apop5:
I'm glad it worked.

I'm going to close this PR for now. If we need to reconsider this change, we can reopen it.

PR #396, opened 2025-01-23
ReleaseWorkflow.yml: Create PR to update version

Javagedes:
Updates the release workflow to trigger on a published release event and create a PR that updates the version of the crate(s) in the workspace instead of pushing a commit directly to the default branch.

### DIFF
--- a/.sync/workflows/leaf/publish-release.yml
+++ b/.sync/workflows/leaf/publish-release.yml
@@ -10,30 +10,16 @@
 name: Publish Release
 
 on:
-  workflow_dispatch:
+  release:
+    types: [published]
 
 jobs:
-  validate_branch:
-    name: Validate Branch
-    runs-on: ubuntu-latest
-
-    steps:
-      - name: Checkout Repository
-        uses: actions/checkout@v4
-
-      - name: Validate Branch
-        run: |
-          if [ "${GITHUB_REF}" != "refs/heads/main" ]; then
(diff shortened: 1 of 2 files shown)


apop5:
Were you able to test run this?

.sync/workflows/leaf/publish-release.yml
makubacki, line 1: Can you update the file description to reflect that the workflow no longer creates a release on GitHub but runs when a release is published?

makubacki:
> > Were you able to test run this?
> 
> @apop5 Yes, this is actively being used on a private repo and is working :)

Have/can you double check it with a test on a repo in the microsoft org? I'm assuming you enabled similar permissions, but this would double check nothing was missed.

Javagedes:
Going to close for now. I don't think this is the best route after seeing the issues with binstall on linux systems.

Javagedes:
No, I need to get back to this.

makubacki:
> No, I need to get back to this. It is not going to work well because we use CRLF line endings where as the cargo-release tool expects LF.

I started something over here at one point too https://github.com/microsoft/mu_devops/pull/412. If you're busy with SEA stuff, I can come back to this.

Javagedes:
Yes I'm pretty busy with SEA stuff. If you want to get this in quickly please take it, otherwise I'll get to it once I've finished up all my TODOs for SEA.

PR #399, opened 2025-02-04
Adding backport labels to filesync, and dependabot PRs

apop5:
Filesync and Dependabot are creating Prs for the dev branches.

This allows dev to become out of sync with release branches.

Adding Backport labels to the PRs generated by filesync and dependabot.

### DIFF
--- a/.sync/FilesBackPort.yml
+++ b/.sync/FilesBackPort.yml
@@ -35,28 +35,11 @@ group:
     repos: |
       microsoft/mu_basecore
       microsoft/mu_common_intel_min_platform
-      microsoft/mu_feature_debugger
-      microsoft/mu_feature_dfci
       microsoft/mu_plus
       microsoft/mu_silicon_arm_tiano
       microsoft/mu_silicon_intel_tiano
-      microsoft/mu_tiano_platforms
       microsoft/mu_tiano_plus
-      microsoft/mu_feature_config
       microsoft/mu_oem_sample
-      microsoft/mu_feature_ffa
-      microsoft/mu_feature_ipmi
-      microsoft/mu_feature_mm_supv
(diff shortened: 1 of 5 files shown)


.github/dependabot.yml
```diff
@@ -39,6 +39,7 @@ updates:
     labels:
       - "type:dependencies"
       - "type:dependabot"
+      - "type:backport"
```
apop5, line 42: @makubacki 

A backport label in a repo that doesn't have the dev/release split should not pose a problem, correct?

makubacki: It won't cause scripts or anything to act on the label, but it would be confusing to a user.

You'd need to update the synced template files anyway in [.sync/dependabot](https://github.com/microsoft/mu_devops/tree/main/.sync/dependabot). So, you could make those be Nunjucks templates and provide an optional parameter to add this label. And then break out the repos that have the dev/release branch split in `.sync/Files.yml` to include it during the sync.

makubacki: @apop5, are you planning to follow up with this change?

apop5: Yes, just trying to finish up a couple of other things and will get back to this.

Javagedes (approved):

makubacki:
@apop5, I'll review when it comes out of draft.

.sync/FilesBackPort.yml
makubacki, line 1: @apop5, I thought the plan was to keep file sync logic (relatively) simple and add a workflow in each repo that needs file syncs backported to trigger on file sync PR creation and add the label. Was there an issue with that approach?

apop5:
closing in favor of #434

PR #407, opened 2025-02-06
Containers/Ubuntu-24: Allow CI user access to venv

makubacki:
Many of the existing CI workflows expect a pip environment to use. A new user and group are created for the docker user when the container image is initialized in pipelines. This allows that user to use the /opt/venv environment to support existing workflows.


makubacki:
Only tested using the container locally before. This is this change in CI - [[TEST] Ubuntu 24.04 Container CI Test](https://github.com/microsoft/mu_tiano_platforms/pull/1108).

apop5 (approved):

Javagedes (approved):

PR #409, opened 2025-02-07
Add Cargo Tool upload pipelines

Javagedes:
Adds a template pipeline that compiles a specified cargo tool for the following target tripples:

1. x86_64-pc-windows-msvc (if `upload_windows == true`)
2. aarch64-pc-windows-msvc (if `upload_windows == true`)
3. x86_64-unknown-linux-gnu (if `upload_linux == true`)
4. aarch64-unknown-linux-gnu (if `upload_linux == true`)

The compiled binaries are formatted in an edk2-pytool-extension external dependency formatted layout.

Adds two consumers of the pipeline for cargo-make and cargo-tarpaulin

### DIFF
--- a/.azurepipelines/CargoMakeArtifact.yml
+++ b/.azurepipelines/CargoMakeArtifact.yml
@@ -13,7 +13,7 @@ parameters:
   - name: python_version
     displayName: 'Python Version'
     type: string
-    default: '3.11'
+    default: '3.12'
 
 extends:
   template: templates/CargoToolRepublish.yml
--- a/.azurepipelines/CargoMakeArtifact.yml
+++ b/.azurepipelines/CargoMakeArtifact.yml
@@ -1,0 +1,26 @@
+## @file
+# Azure Pipeline to build and publish Cargo Make in a format supported by edk2-pytool-extension's external dependency.
+#
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+trigger: none
+
+parameters:
+  - name: tag
+    displayName: 'Cargo Make Version Tag'
+    type: string
+  - name: python_version
+    displayName: 'Python Version'
+    type: string
+    default: '3.11'
+
+extends:
+  template: templates/CargoToolRepublish.yml
+  parameters:
+    tool_name: 'cargo-make'
+    tag: ${{ parameters.tag }}
+    repo_url: 'https://github.com/sagiegurari/cargo-make.git'
+    python_version: ${{ parameters.python_version }}
+    upload_linux: true
+    upload_windows: true
--- a/.azurepipelines/CargoTarpaulinArtifact.yml
+++ b/.azurepipelines/CargoTarpaulinArtifact.yml
@@ -14,7 +14,7 @@ parameters:
   - name: python_version
     displayName: 'Python Version'
     type: string
-    default: '3.11'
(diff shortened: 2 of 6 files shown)


.azurepipelines/CargoMakeArtifact.yml
```diff
@@ -0,0 +1,26 @@
...
+trigger: none
+
+parameters:
+  - name: tag
+    displayName: 'Cargo Make Version Tag'
+    type: string
+  - name: python_version
+    displayName: 'Python Version'
+    type: string
+    default: '3.11'
```
makubacki, line 16: Not a big deal, but we're using 3.12 most other places so it might be nice to start there.

.azurepipelines/templates/CargoToolRepublish.yml
```diff
@@ -0,0 +1,194 @@
...
+  - name: python_version
+    displayName: 'Python Version'
+    type: string
+    default: '3.11'
+
+variables:
+  - name: windows_image
+    value: 'windows-latest'
+  - name: linux_image
+    value: 'ubuntu-22.04'
```
makubacki, line 41: Please use Ubuntu 24.04 since we've updated to that everywhere else.

.azurepipelines/templates/CargoToolRepublish.yml
```diff
@@ -0,0 +1,194 @@
...
+
+stages:
+- stage: build_binaries
+  displayName: 'Build ${{ parameters.tool_name }} Artifact(s)'
+
+  jobs:
+    - job: build_linux
+      displayName: Build Linux Binaries
```
makubacki, line 49: It looks like these Linux & Windows build sections might be a little cleaner with a matrix from aarch64 and x86_64. But there's not much as-is either.

Javagedes: Things are different enough between the two pipelines that I decided to have two separate jobs rather than a pipeline. 

There differences are mainly:

1. different architectures downloaded per OS
2. Linux uses `cross` to compile to reduce complexity of cross compilation on linux
3. linux makes the binaries execute

I can look into making it a matrix if you want, but it may not be worth the time it will take to make it work

.azurepipelines/templates/CargoToolRepublish.yml
Javagedes: switched to 24.04

.azurepipelines/CargoMakeArtifact.yml
Javagedes: switched to 3.12

.azurepipelines/templates/CargoToolRepublish.yml
makubacki: I think it's okay as-is. It's not something I'd expect to benefit tremendously from the matrix anyway in terms of future extension.

.azurepipelines/CargoTarpaulinArtifact.yml
```diff
@@ -0,0 +1,28 @@
...
+trigger: none
+
+parameters:
+  - name: tag
+    displayName: 'Cargo Tarpaulin Version Tag'
+    type: string
+  - name: python_version
+    displayName: 'Python Version'
+    type: string
+    default: '3.11'
```
makubacki, line 17: Need 3.12 here too.

makubacki (approved):

PR #412, opened 2025-02-21
.github/ReleaseWorkflow.yml: Push version to branch

makubacki:
- Use `--no--tag` on dry run to prevent failure if tag already exists on the GitHub repo
- Create and push branch with the version update and make link available in GitHub workflow output.


makubacki:
Need to test on fork. After that is done, will move out of draft.

PR #415, opened 2025-03-20
Rust toolchain: Update to 1.84

antklein:
Update MU to Rust toolchain 1.84.

### DIFF
--- a/.sync/Version.njk
+++ b/.sync/Version.njk
@@ -43,7 +43,7 @@
 {% set python_version = "3.12" %}
 
 {# The Rust toolchain version to use. #}
-{% set rust_toolchain = "1.80.0" %}
+{% set rust_toolchain = "1.84.0" %}
 
 {# Rust tool versions. #}
 {% set cargo_make = "0.37.24" %}


makubacki:
@antklein, are you planning to make any changes in Mu repos that are needed?

antklein:
> @antklein, are you planning to make any changes in Mu repos that are needed?

I can. :)
This was just a hopeful PR to align the toolchain with other projects. So far, I have done zero validation for repositories that would be updated via this change.

I'll validate these repositories before completing this PR.
      microsoft/mu_basecore
      microsoft/mu_plus
      microsoft/mu_rust_helpers
      microsoft/mu_rust_hid
      microsoft/mu_rust_pi
      microsoft/mu_tiano_platforms

antklein:
> > @antklein, are you planning to make any changes in Mu repos that are needed?
> 
> I can. :) This was just a hopeful PR to align the toolchain with other projects. So far, I have done zero validation for repositories that would be updated via this change.
> 
> I'll validate these repositories before completing this PR. microsoft/mu_basecore microsoft/mu_plus microsoft/mu_rust_helpers microsoft/mu_rust_hid microsoft/mu_rust_pi microsoft/mu_tiano_platforms

@makubacki I validated cargo `debug` and `release` builds all pass for the above repositories with rust toolchain 1.84.

Javagedes:
@antklein builds typically do not break while updating cargo to new versions. I suspect where you will see breakage is with cargo clippy or cargo fmt. Rust is pretty good about backwards compatibility. The only place that is not guaranteed are in our unsafe feature usage.

It looks like mu_devops steps do not actually run clippy; I don't suspect this will cause any breakage, though there is a chance with cargo fmt.

antklein:
> > @antklein builds typically do not break while updating cargo to new versions. I suspect where you will see breakage is with cargo clippy or cargo fmt. Rust is pretty good about backwards compatibility. The only place that is not guaranteed are in our unsafe feature usage.
> > It looks like mu_devops steps do not actually run clippy; I don't suspect this will cause any breakage, though there is a chance with cargo fmt.
> 
> To clarify my earlier request, I meant all of our common cargo commands - clippy, fmt, build, doc, test. CI is one thing (and can be updated), but I also don't want the first developer to run the command having to fix problems.

Thanks for the clarification Michael and Joey. I have gone through the repositories and verified that moving to 1.84 causes no new clippy/fmt errors. All tests are passing.

PR #438, opened 2025-05-05
Cherry Pick Workflow To Ensure Consistent Cherry-Picking

Flickdm:
## Description

This workflow ensures that cherry-picks are consistent and retain the relevant history for integrations.

Any PR with [Cherry Pick] or [Cherry-Pick] will trigger this workflow

It will check the following:
 1. Cherry picks contain "[CHERRY-PICK/*] *"
 2. (cherry picked from commit <sha256>)

If it's a "Merge" commit the commit will be ignored. (Needed during PR)

If the commit violates one of these policies, then the PR gate will fail with the failing commit

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?
- [ ] Backport to release branch?

## How This Was Tested

https://github.com/Flickdm/mu_tiano_plus/pull/1

## Integration Instructions

N/A

### DIFF
--- a/.github/workflows/cherry-pick.yml
+++ b/.github/workflows/cherry-pick.yml
@@ -1,6 +1,6 @@
 # This workflow validates commit messages in a pull request to ensure they follow a specific pattern.
 # The pattern is defined as follows:
-# - The commit message must contain the string "[cherry-pick/XXX]" where XXX is a valid label.
+# - The commit message must contain the string "[cherry-pick]".
 # - The commit message must also contain "(cherry picked from commit <commit_hash>)" where <commit_hash> is a 40-character SHA-1 hash.
 # - Merge commits are ignored.
 #
@@ -20,7 +20,7 @@ on:
 
 jobs:
   cherry-pick-job:
-    if: contains(github.event.pull_request.title, '[cherry pick]') || contains(github.event.pull_request.title, '[cherry-pick]')
+    if: contains(github.event.pull_request.labels.*.name, 'cherry-pick')
     runs-on: ubuntu-latest
 
     steps:
@@ -35,7 +35,7 @@ jobs:
 
         # Ensure the base branch is fetched
         git fetch origin $base_branch
-        
+
         # Get the commits included in the PR
         commits=$(git log --pretty=format:%H origin/$base_branch..HEAD)
 
@@ -49,7 +49,7 @@ jobs:
               continue
             fi
 
-            if [[ ! $(echo "$commit_message" | tr '[:upper:]' '[:lower:]') =~ \[cherry-pick\/.+\] ]] || [[ ! "$commit_message" =~ \(cherry\ picked\ from\ commit\ [0-9a-f]{40}\) ]]; then
+            if [[ ! $(echo "$commit_message" | tr '[:upper:]' '[:lower:]') =~ \[cherry-pick\] ]] || [[ ! "$commit_message" =~ \(cherry\ picked\ from\ commit\ [0-9a-f]{40}\) ]]; then
               echo "Commit $commit does not meet the required pattern."
               echo "Commit message: $commit_message"
               echo "Please use 'git cherry-pick -x <commit_hash>' to cherry-pick the commit."
--- a/.github/workflows/cherry-pick.yml
+++ b/.github/workflows/cherry-pick.yml
@@ -1,0 +1,62 @@
+# This workflow validates commit messages in a pull request to ensure they follow a specific pattern.
+# The pattern is defined as follows:
+# - The commit message must contain the string "[cherry-pick/XXX]" where XXX is a valid label.
+# - The commit message must also contain "(cherry picked from commit <commit_hash>)" where <commit_hash> is a 40-character SHA-1 hash.
+# - Merge commits are ignored.
+#
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+
+name: Cherry-Pick Workflow
+
+on:
+  pull_request:
+    types:
+      - edited
+      - opened
+      - reopened
+      - synchronize
+  workflow_dispatch:
+
+jobs:
+  cherry-pick-job:
+    if: contains(github.event.pull_request.title, '[cherry pick]') || contains(github.event.pull_request.title, '[cherry-pick]')
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0  # Fetch the full history to ensure all commits are available
+
+    - name: Validate commit messages
+      run: |
+        base_branch="${{ github.event.pull_request.base.ref }}"
+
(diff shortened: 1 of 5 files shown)


.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,62 @@
...
+    types:
+      - edited
+      - opened
+      - reopened
+      - synchronize
+  workflow_dispatch:
+
+jobs:
+  cherry-pick-job:
+    if: contains(github.event.pull_request.title, '[cherry pick]') || contains(github.event.pull_request.title, '[cherry-pick]')
```
os-d, line 23: This is case insensitive?

.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,62 @@
+# This workflow validates commit messages in a pull request to ensure they follow a specific pattern.
+# The pattern is defined as follows:
+# - The commit message must contain the string "[cherry-pick/XXX]" where XXX is a valid label.
```
os-d, line 3: I'm a bit confused on this requirement, what does this provide? Is this something `git cherry-pick -x` does?

.github/workflows/cherry-pick.yml
Flickdm: Yes

.github/workflows/cherry-pick.yml
Flickdm: No we would need to rebase and add [cherry-pick] cherry-pick -x simply adds the history comment at the base of the commit message

.github/workflows/cherry-pick.yml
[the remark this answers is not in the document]
Flickdm: Honestly I would like a label but then we would need to identify a thing to be labeled and run into the same problem but I agree this is more brittle than I would like. Maybe we move this logic to the labler and then check for the existence of the label here?

.github/workflows/cherry-pick.yml
os-d: Oh oh, I see. I was missing this was intended to be the subject of the commit message. I thought it was trying to enforce this in the body, which I didn't see why that added value over having the line about what commit it was cherry-picked from. I think Cherry-Pick in the commit subject makes sense (though only works if this is a rebase). What are you envisioning the XXX would be? Is `[Cherry-Pick]` not enough?

.github/workflows/cherry-pick.yml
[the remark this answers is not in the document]
Flickdm: > This `if` seems pretty finicky to me. because someone could mess this up in many different ways with capitlization and such.
> 
> I see two good ways to solve this:
> 
> 1. Always run this, but have the first step run validation (so maybe lowercase this, remove any "-", "_", or " " in the commit message, and see if [cherrypick] is contained. Produce a variable that is used in the `if` check of all other steps
> 2. Base this off of a label being added or removed with the trigger `labeled`. with the `if` statement being `if: github.event.label.name == 'cherry-pick'`
> 
> With (2) we could also have a check box in the PR for marking something as a cherry-pick. If it is a cherry-pick the auto-labeler will add the cherry-pick label, which would then cause this action to fire.

I see you edited your comment - a new box for a "cherry-pick" could be a good idea

.github/workflows/cherry-pick.yml
makubacki, line 1: Can this just be added to [Pull Request Formatting Validator](https://github.com/microsoft/mu_devops/blob/main/.github/workflows/pull-request-formatting-validator.yml)?

.github/workflows/cherry-pick.yml
Flickdm: My example right now is `[CHERRY-PICK/EDK2]` to signify where it was cherry-picked from however I'm willing to remove it. I wasn't sure if the source part was very useful

.github/workflows/cherry-pick.yml
Flickdm: Do you think it makes sense to add this to the general pull request formatter? I'm happy to add it there if you think it's worth it. I was just trying to create it as a separate workflow to separate the logic

.github/workflows/cherry-pick.yml
Javagedes: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#contains

.github/workflows/cherry-pick.yml
os-d: We only have one upstream, so I don't think it is of much value. Maybe for tracking between Mu branches, but I would be of the opinion to just have `[CHERRY-PICK]` in the commit subject. The real goal for this, as I see it, is for the integrator of the next edk2 release to be able to easily drop the cherry-picks and reference them for the reverse operation (cherry-picking a potentially conflicting MU_CHANGE with what edk2 has).

.github/workflows/cherry-pick.yml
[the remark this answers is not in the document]
Javagedes: As You commented on Olivers message, you confirmed `contains` is not case sensitive, so this is less of a concern for me. I would maybe just add cherry_pick as a possible option, worst case.

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#contains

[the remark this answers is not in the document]
makubacki: If we want this, I suggest the checkbox + label. Then, it is a straightforward opt-in with the label attached for further filtering versus the opportunity for typos and/or misunderstanding what the tag in the PR title means.

.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,62 @@
...
+
+        for commit in $commits; do
+            commit_message=$(git log --format=%B -n 1 "$commit")
+
+            if [[ "$commit_message" =~ ^Merge ]]; then
+              echo "Warning: Commit $commit is a merge commit and will be ignored."
+              continue
+            fi
+
+            if [[ ! $(echo "$commit_message" | tr '[:upper:]' '[:lower:]') =~ \[cherry-pick\/.+\] ]] || [[ ! "$commit_message" =~ \(cherry\ picked\ from\ commit\ [0-9a-f]{40}\) ]]; then
```
makubacki, line 52: The user visible messages are going to need to be more visible in the PR (comments). Many people don't know to look at actions log messages.

.github/workflows/cherry-pick.yml
makubacki: That's meant to be a collection of checks for formatting. You can still conditionalize this on a "cherry-pick trigger".

.github/workflows/cherry-pick.yml
makubacki: I agree with dropping the cherry-pick source for simplicity.

makubacki:
Closing for now. @Flickdm agrees.

PR #444, opened 2025-05-28
Update rust to 1.85.0

Javagedes:
Updates the rust version to 1.85.0


makubacki (approved):

Javagedes:
> Is this not considered a breaking change because you're going to make the Mu repo updates needed?

Sure, I'd say you can consider it a breaking change. Will mark it as such and update the version in my change.

apop5 (approved):

PR #446, opened 2025-05-29
Update rust version in setup steps

Javagedes:
As is always the case with mu_devops, I missed something.

The nice thing is it does not look like we need to do an immediate release or file sync. The three rust repos all ride `main`, so a re-run once this is merged should hopefully suffice.


makubacki:
Is https://github.com/microsoft/mu_devops/pull/445 sufficient?

PR #458, opened 2025-06-27
Use the GitHub app token for writing

makubacki:
Remove uses of the default GitHub token with write permission and instead use the GitHub app derived token which has write access.

- Removes setting the default token to have write access.
- Uses `pull_request_target` for the pull request triggered workflow to allow the secrets to be passed to the action used to derive the GitHub app token.


makubacki:
I don't think the PR status check will complete successfully until this in the `main` branch.

Issue #459, opened 2025-06-20
[Task] Follow up on CodeQL workflow `cargo make` failure


makubacki:
The CodeQL workflow in mu_plus dev/202502 began failing for PRs. Example: https://github.com/microsoft/mu_plus/actions/runs/15787129446

A quick change was made to unblock other PRs: https://github.com/microsoft/mu_plus/pull/705

That needs to be followed up on for a root cause and longer term fix.

PR #460, opened 2025-07-02
codeql: Use app token for API requests [Rebase & FF]

makubacki:
Two changes for the CodeQL workflows:

---

**codeql: Use app token for API requests**

Make authenticated requests to prevent relying on the GitHub
anonymous API limit from potentially causing requests to
fail.

---

**codeql: Always download cargo make**

Temporarily always download cargo make instead of using the workflow
cache as loading from the cache has failed recently in some repos
and a root cause needs to be found for that issue.

Tracked in https://github.com/microsoft/mu_devops/issues/459

### DIFF
--- a/.sync/workflows/leaf/codeql-platform.yml
+++ b/.sync/workflows/leaf/codeql-platform.yml
@@ -34,7 +34,7 @@ on:
       - main
       - release/*
       - dev/*
-  pull_request:
+  pull_request_target:
     branches:
       - main
       - release/*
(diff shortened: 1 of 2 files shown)


Javagedes (approved):

makubacki:
This is another PR I'd like to double check before completing in mu_devops. Created this mu_basecore PR to do that https://github.com/microsoft/mu_basecore/pull/1421.

PR #465, opened 2025-07-16
Add to the dependenabot ignore subrepos the TFA and HAF repos

apop5:
mu_tiano_platforms added Silicon/Arm/TFA and Silicon/Arm/HAF as submodules.

Dependabot was creating PRs to update.

Add these repos to the ignore list for dependabot.

### DIFF
--- a/.sync/dependabot/actions-pip-submodules.yml
+++ b/.sync/dependabot/actions-pip-submodules.yml
@@ -87,6 +87,8 @@ updates:
       - dependency-name: "MU_BASECORE"
       - dependency-name: "Silicon/Arm/MU_TIANO"
       - dependency-name: "Silicon/Intel/MU_TIANO"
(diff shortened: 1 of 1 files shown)


os-d (approved):
I'm confused, is this ignore list saying dependabot isn't doing any of our submodules? Is that because of some break?

apop5:
Dependabot isn't supposed to handle our submodules. Dependabot doesn't have the capability to update to tags, only to latest commits.

submodules release updater (https://github.com/microsoft/mu_devops/tree/main/.github/actions/submodule-release-updater)
is supposed to handle out updates of submodules.

adding the tfa and haf submodules in mu tiano platforms, without updating the ignore dependency, resulted in depentabot attempting to handle submodule updates.  This will just prevent dependabot from being able to update those

Javagedes (approved):

kuqin12 (approved):

PR #466, opened 2025-07-18
Update rustfmt to use local repo version instead of mu_devops version.

apop5:
Some downstream repos updated to 2024 rustfmt.

Update the synced version of rustfmt to remove specifying a format.  This way, the edition will use the repo's Cargo.toml.

### DIFF
--- a/.sync/rust_config/rustfmt.toml
+++ b/.sync/rust_config/rustfmt.toml
@@ -5,7 +5,7 @@
 
 # Keep these options sorted in ascending order to ease lookup with rustfmt documentation.
 
-edition = "2021"                 # This would normally be picked up from Cargo.toml if not specified here
+edition = "2024"                 # This would normally be picked up from Cargo.toml if not specified here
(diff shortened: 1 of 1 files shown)


.sync/rust_config/rustfmt.toml
```diff
@@ -5,7 +5,7 @@
 
 # Keep these options sorted in ascending order to ease lookup with rustfmt documentation.
 
-edition = "2021"                 # This would normally be picked up from Cargo.toml if not specified here
+edition = "2024"                 # This would normally be picked up from Cargo.toml if not specified here
```
makubacki, line 8: Let's just delete it to keep things simple.

```suggestion
```

makubacki: It will fall back to Cargo.toml. Which I think is fine for Mu repos. We don't have so much Mu Rust code that tracking fmt edition updates separately seems necessary.

makubacki (approved):

PR #478, opened 2025-08-27
Personal/apop5/rustci

apop5:
Makefile.toml - Check if default-members is empty before attempting to run clippy.

Update Jobs/PrGate to run RustCargoSteps after the repo's dependencies have been fully hydrated.


makubacki:
> @makubacki This issue comes about from platforms that have Rust EFI's defined in submodules. These cannot be defined as git dependencies because I cannot run `cargo build` on an external crate, only a crate defined in the workspace.
> 
> You can theoretically do a `cargo install` on a git dependency, but that would require changes to the build rules template.

You're right, this was more about the library crates. Since the only binary dependencies are the Rust Hello World driver and the HID driver, we can better achieve our overall goals to simplify mu_tiano_platforms usage and reduce overall complexity of supporting Rust alongside the EFI build system by removing the modules from mu_tiano_platforms and dropping the Rust files from being synced there. As discussed offline, mu_tiano_platforms can use the C HID driver.

Javagedes (approved):

makubacki (approved):

PR #498, opened 2025-09-22
CodeQL workflows: Remove dev branch trigger

makubacki:
Does not trigger the workflow on the dev branch since the dev branch concept has been dropped from Project Mu.

---

One more reference found that is not needed.


makubacki:
Plan is to amend open sync PRs with this final change.

PR #568, opened 2026-05-01
Update Version.njk mu_devops to 18.0.6

Flickdm:
Updates the Version.njk mu_devops  to 18.0.6

PR #569, opened 2026-05-01
Revert "Update Version.njk mu_devops to 18.0.6"

Flickdm:
Reverts microsoft/mu_devops#568


makubacki:
Let's see if this is needed. If the CodeQL workflow is fine in mu_crypto_release we can proceed with what's checked in.

PR #580, opened 2026-06-10
Handle “no new commits” scenario in GenerateTag template as a successful no‑op

danivergaurav:
### Summary

Handle the “no new commits since last tag” scenario in `GenerateTag.yml` as a successful no‑op instead of failing the pipeline.

### Problem

When no new commits exist:

- `TagGenerator.py` exits successfully but does not set `tag_name`
- Pipeline still attempts to create/push a tag
- Results in failure (tag_name: command not found)

  
Failure_No_Commit

### Fix

- Detect absence of `tag_name` after `TagGenerator` execution
- Introduce `skip_tag_generation` flag
- Skip tag/commit/push steps when no tag is generated
- Add a no‑op completion message

### Result

- Pipeline succeeds when there are no changes, it has been tested in ADO.
- Tag is created only when needed
- No masking of real failures

Successful_run

### DIFF
--- a/Jobs/GenerateTag.yml
+++ b/Jobs/GenerateTag.yml
@@ -75,6 +75,16 @@ jobs:
         displayName: "Run Tag Generator"
         workingDirectory: $(Agent.BuildDirectory)
 
+      - bash: |
+          echo "Generated tag: '$(tag_name)'"
+          if [ -z "$(tag_name)" ]; then
+          echo "No new commits since last tag."
+            echo "##vso[task.setvariable variable=skip_tag_generation]true"
+          else
+            echo "##vso[task.setvariable variable=skip_tag_generation]false"
+          fi
+        displayName: Detect no-op tag generation
+        
       - script: |
           set -e
           git branch
@@ -86,3 +96,4 @@ jobs:
         continueOnError: false
         displayName: "Create Tag"
         workingDirectory: $(Agent.BuildDirectory)/target
+        condition: ne(variables.skip_tag_generation, 'true')


apop5 (approved):

Jobs/GenerateTag.yml
```diff
@@ -75,6 +75,16 @@ jobs:
         displayName: "Run Tag Generator"
         workingDirectory: $(Agent.BuildDirectory)
 
+      - bash: |
+          echo "Generated tag: '$(tag_name)'"
```
cfernald, line 79: Is there a way we could directly detect this from TagGenerator.py instead? Having the heuristic of the variable not being set seems like it may cause us to assume the reason incorrectly in the future.

Javagedes (approved):

Jobs/GenerateTag.yml
```diff
@@ -75,6 +75,16 @@ jobs:
...
         workingDirectory: $(Agent.BuildDirectory)
 
+      - bash: |
+          echo "Generated tag: '$(tag_name)'"
+          if [ -z "$(tag_name)" ]; then
+          echo "No new commits since last tag."
+            echo "##vso[task.setvariable variable=skip_tag_generation]true"
+          else
+            echo "##vso[task.setvariable variable=skip_tag_generation]false"
+          fi
```
makubacki, line 85: If this task is kept, can Line 81 be aligned?

```suggestion
          if [ -z "$(tag_name)" ]; then
            echo "No new commits since last tag."
            echo "##vso[task.setvariable variable=skip_tag_generation]true"
          else
            echo "##vso[task.setvariable variable=skip_tag_generation]false"
          fi
```

Jobs/GenerateTag.yml
makubacki: I agree that would be cleaner to more explicitly handle this in `TagGenerator.py`.

apop5:
Based on offline conversation, feedback will be implemented at a later point in time, when resources are freed up

makubacki:
> Based on offline conversation, feedback will be implemented at a later point in time, when resources are freed up

Can we close this PR for now then? It can always be reopened. It makes it easier to focus on what is active and needs review.

danivergaurav:
As suggested, closing this PR for now.

PR #581, opened 2026-06-12
Update Ubuntu containers to install QEMU v11 and swtpm

kuqin12:
This change updates the container image script to install QEMU v11.

This change also installs `swtpm` rather than requiring the consumer pipeline to install it separately.


makubacki (approved):

vineelko (approved):

kuqin12:
Tested the image locally and was able to verify the qemu version and swtpm is also installed.

PR #596, opened 2026-07-27
Dependabot: Add 7 day cooldown period to dependabot updates.

apop5:
This is to improve supply chain security by giving a seven day
buffer for malicious dependencies to be reported before
Mu repos/platform adopt the dependency.


makubacki (approved):
@apop5, I think you could mention in the commit/PR description that this is improves supply chain security by giving a buffer of time for malicious dependencies to be reported before Mu platform consumption.

Javagedes (approved):

PR #602, opened 2026-09-02
Containers: Install 32-bit development packages for IA32 host tests

apop5:
Add the 32-bit libc development packages to the Ubuntu build containers.

The Host Unit Test Compiler Plugin builds IA32 applications with -m32.
Without the 32 bit development headers, C++ standard library includes
eventually fail in linux/errno.h because asm/errno.h cannot be found.

Installing these packages provides the required IA32 headers, runtime
libraries, startup objects, and compiler support.

Created local docker image for ubuntu-24 and verified that IA32 GCC NOOPT builds completed successfully.


apop5:
Failures can be found here:

https://github.com/microsoft/mu_basecore/pull/1901

Javagedes (approved):

makubacki (approved):

